PR #9, opened 2015-10-19
Close #8: Get_fields returns extends in different order than the field array

### LINKED ISSUE #8: Get_fields returns extents in different order than the field array
I noticed an unexpected behaviour in how `get_fields` returns the fields and the extent of the fields: 
while the extents are returned as (z_min, z_max, r_min, r_max), or similar for other coordinates, the shape of the field array is (Nr, Nz). In other words "flipped" in order. I think this behaviour can lead to confusion and we should, if we don't change this, at least explicitly mention the structure of the returned quantities in the docstring.

RemiLehe:
This was solved by returning a FieldMetaInformation object, instead of the simple array `extent`.

The plotter and the tutorial notebook have been updated accordingly. The tutorials do not give a lot of details on the FieldMetaInformation object, but the docstring of the object tells more about the available attributes. 

NB : The docstring of the FieldMetaInformation's `__init__` is very short, which goes against the recommendations in CONTRIBUTING.md. The reason is so that users can concentrate on the docstring of the class itsself when doing `FieldMetaInformation?` in an IPython environnement, instead of the docstring of the `__init__` function which is much less useful for a user.


soerenjalas:
This is really elegant solution to the whole extent problem. Thanks!

RemiLehe:
Thanks for reviewing this. Many thanks also to @PrometheusPi for suggesting that solution. 

I'll merge it (only in `dev` for the moment). Please pull these changes to the branch where you are implementing the laser-plasma diagnostic. I guess implementing the laser-plasma diagnostics will be a great way to test how practical the FieldMetaInformation is. If you need additional metainformation, let me know ; I can always add them to the object.

Once this has been used and tested for a while, we can push that to the master branch. 

Remi

PR #10, opened 2015-10-20
Implementation of Lpa Diagnostic class

soerenjalas:
I wrote the LPA diagnostic class with refined versions of all diagnostic methods from the early version of the viewer. I realise that this file is rather long and splitting it up might be an option (e.g particle analysis and field analysis), but in my eyes this tends to make the code more complicated. I tried to comment the code thoroughly. However, if something is not clear to you please don't hesitate to ask. 
The particle methods are prepared for selection via a selection dictionary.


RemiLehe:
Great, thanks a lot for the pull request !!
I am already quite a fan of `get_envelope` and `get_wigner`, and I look forward to using them in production!

Before merging, I had a few questions and suggestions (see comments in the code). Please let me know if you agree with them. Also, if you don't have to integrate them in the code, please let me know, and I can do it myself (as long as I know that you agree with them)

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
             one file per iteration, and the name of the files should end
             with the iteration number, followed by '.h5' (e.g. data0005000.h5)
         """
-        OpenPMDTimeSeries.__init__( path_to_dir )
+        super(LpaDiagnostics, self).__init__( path_to_dir )
```
RemiLehe: What about doing
```python
OpenPMDTimeSeries.__init__( self, path_to_dir )
```
instead of 
```python
super(LpaDiagnostics, self).__init__( path_to_dir )
```
as it is more transparent. 

(I forgot the `self` in the previous version of the code; sorry)

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+
+        Returns
+        -------
+        A tuple of floats with:
+        - mean weighted gamma
+        - weighted standard deviation of gamma
+        """
+        # Find the output that corresponds to the requested time/iteration
+        # (Modifies self.current_i and self.current_t)
+        self._find_output( t, iteration )
```
RemiLehe: Actually, `self._find_output( t, iteration )` is already called in the following `self.get_particle` (or `self.get_field` in other functions), so I would suggest to remove the paragraph
```python
 # Find the output that corresponds to the requested time/iteration
 # (Modifies self.current_i and self.current_t)
 self._find_output( t, iteration )
```
from this function and the following ones.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        # Get particle data
+        ux, uy, uz, w = self.get_particle(
+                         var_list=['ux', 'uy', 'uz', 'w'],
+                         species=species, t=t, iteration=iteration )
+        # Calculate Lorentz factor for all particles
+        gamma = np.sqrt(1 + ux ** 2 + uy ** 2 + uz ** 2)
+        # Calculate weighted mean and average
+        try:
+            # Calculate mean_gamma for selected particles
+            mean_gamma = np.sqrt(np.average(gamma ** 2, weights=w))
```
RemiLehe: I was wondering if there is a particular reason to calculate the RMS of gamma here, instead of the actual mean (i.e. `np.average(gamma, weights=w)`) - which would maybe be more standard.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        A float with the electric charge of the selected particles in Coulomb
+        """
+        # Find the output that corresponds to the requested time/iteration
+        # (Modifies self.current_i and self.current_t)
+        self._find_output( t, iteration )
+        # Get particle data
+        w, q = self.get_particle( var_list=['w', 'charge'], species=species,
+                                 t=t, iteration=iteration )
+        # Calculate charge
+        charge = np.sum(w) * q
```
RemiLehe: One thing to keep in mind is that `q` could in principle be an array (with size = the number of macroparticles) with different values of the charge (for instance, a bunch of ions that have been ionized to various degrees) - as openPMD allows for that kind of thing. For this reason, I would replace
```python
charge = np.sum(w) * q
```
by
```python
charge = np.sum(w * q)
```
That way, the final result is bound to be a scalar, not an array. 
Is that okay with you ?

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        # (Modifies self.current_i and self.current_t)
+        self._find_output( t, iteration )
+        # Get particle data
+        ux, uy, uz, w = self.get_particle( var_list=['ux', 'uy', 'uz', 'w'],
+                                           t=t, iteration=iteration,
+                                           species=species )
+        # Calculate diveregence
+        div_x = wstd( ux / uz, w )
+        div_y = wstd( uy / uz, w )
+        # Return the result
```
RemiLehe: What about doing
```python
div_x = wstd( np.arctan2(ux, uz), w )
div_y = wstd( np.arctan2(ux, uz), w )
```
so that it is also valid for large deflection angles, and does not crash for uz=0.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+            Either None or a dictionary of rules
+            to select the particles, of the form
+            'x' : [-4., 10.]   (Particles having x between -4 and 10 microns)
+            'z' : [0, 100] (Particles having x between 0 and 100 microns)
+
+        Returns
+        -------
+        A tuple with :
+        - normalized beam emittance in the x plane (pi m rad)
+        - normalized beam get_emittance in the y plane (pi m rad)
```
RemiLehe: One should maybe replace `get_emittance` by `emittance` in the last line ;)

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+                                    species=species )
+        # Calculate the necessary RMS values
+        x *= 1.e-6
+        y *= 1.e-6
+        xsq = np.average( x ** 2, weights=w )
+        ysq = np.average( y ** 2, weights=w )
+        uxsq = np.average( ux ** 2, weights=w )
+        uysq = np.average( uy ** 2, weights=w )
+        xpx = np.average( x * ux, weights=w )
+        ypy = np.average( y * uy, weights=w )
```
RemiLehe: This is really minor, but could you replace `xpx` and `ypy` by `xux` and `yuy` (just for consistency with uxsq) ?
This change would also have to be propagated in the following 2 lines.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        species : string
+            Particle species to use for calculations
+
+        select : dict, optional
+            Either None or a dictionary of rules
+            to select the particles, of the form
+            'x' : [-4., 10.]   (Particles having x between -4 and 10 microns)
+            'z' : [0, 100] (Particles having x between 0 and 100 microns)
+
+        bins : float
```
RemiLehe: `bins : float` should maybe be replaced by `bins : int, optional` :)

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        z, uz, uy, ux, w, q = self.get_particle(
+                               var_list=['z', 'uz', 'uy', 'ux', 'w', 'charge'],
+                               t=t, iteration=iteration,
+                               species=species )
+        # Calculate Lorentz factor for all particles
+        gamma = np.sqrt(1 + ux ** 2 + uy ** 2 + uz ** 2)
+        # Length to be seperated in bins
+        len_z = np.max(z) - np.min(z)
+        # Define bins
+        bins_list = np.linspace(np.min(z), np.max(z), bins)
```
RemiLehe: `np.min(z)` and `np.max(z)` are calculated twice, which is not very efficient (z can be very long, as it represents the particles). What do you think about replacing
```python
 # Length to be seperated in bins
 len_z = np.max(z) - np.min(z)
 # Define bins
 bins_list = np.linspace(np.min(z), np.max(z), bins)
```
by
```python
 # Length to be separated in bins
zmin = np.min(z)
zmax = np.max(z)
 len_z = zmax - zmin
 # Define bins
 bins_list = np.linspace(zmin, zmax, bins)

File: opmd_viewer/addons/pic/lpa_diagnostics.py [hunk 10]
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        # Get bin index of each particle
+        particle_bin = np.digitize(z, bins_list)
+        # Calculate sum of vz in each bin
+        vz_sum = np.zeros_like(bins_list)
+        for i in range(bins):
+            vz_sum[i] = np.sum(uz[particle_bin == i] / gamma[particle_bin == i]
+                               * const.c * w[particle_bin == i] )
```
RemiLehe: It is possible that the above operation is quite inefficient (e.g. if `bins` is 100, then one is searching through the array 100 times). 
I guess that `np.histogram( z, bins=bins, weights = w*uz/gamma )` should return something close to what we need, and is probably more efficient (although I did not check).

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+            an available file, the last file before `t` will be used)
+            Either `t` or `iteration` should be given by the user.
+
+        iteration : int
+            The iteration at which to obtain the data
+            Either `t` or `iteration` should be given by the user.
+
+        pol : float
+            Polarization angel of the field relative to the x plane. Can be
+            freely chosen between 0 and Pi/2 in thetaMode. For carthesian
```
RemiLehe: `carthesian` should probably be replaced by `Cartesian` throughout the text :)

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+            vz_sum[i] = np.sum(uz[particle_bin == i] / gamma[particle_bin == i]
+                               * const.c * w[particle_bin == i] )
+        # Calculete the current in each bin
+        current = vz_sum * np.abs(q) * bins / (len_z * 1.e-6)
+
+        # Return the current and bin edges
+        return(current, bins_list)
+
+    def get_laser_envelope( self, t=None, iteration=None, pol=None, m='all',
+                            freq_filter=40, index='center' ):
```
RemiLehe: This function is great !!!

I have a minor request on the API, though: I find the call for the `y` polarization a little bit cumbersome:
```
ts.get_laser_envelope( pol=np.pi/2 )
```
as it is not very readable and moreover forces people to import numpy.

Therefore, what do you think about modifying the API, so that `pol` is either `x` or `y` (I think these are really the most common case; and if someone really wants a different angle, they can combine it with)
```python 
theta = 0.5 #rad
envx, _ = ts.get_laser_envelope( pol='x' ) 
envy, _ = ts.get_laser_envelope( pol='y' )
env = np.cos(theta)*envx + np.sin(theta)*envy
```
Also, in this way people will write code that works for any geometry (whereas with the current API, selecting an arbitrary angle will only work for thetaMode)

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+                coord = 'y'
+                pol = None
+            else:
+                raise ValueError('Only polarization in the x or y plane is '
+                                 'supported for carthesian coordinates')
+        else:
+            # Extract radial if in thetaMode
+            coord = 'r'
+            slicing_dir = None
+        # Get field data
```
RemiLehe: If you agree with the above change of API, then the folllowing section 
``` python
# Test that polarization is in x or y plane for 3D cart. coordinates
 if self.geometry == "3dcartesian" or self.geometry == "2dcartesian":
 if pol == 0:
 slicing_dir = 'y'
 coord = 'x'
 pol = None
 elif pol == np.pi/2:
 slicing_dir = 'x'
 coord = 'y'
 pol = None
 else:
 raise ValueError('Only polarization in the x or y plane is '
 'supported for carthesian coordinates')
 else:
 # Extract radial if in thetaMode
 coord = 'r'
 slicing_dir = None
 field = self.get_field( t=t, iteration=iteration, field='E',
 coord=coord, theta=pol, m=m,
 slicing_dir=slicing_dir )
```
could maybe be replaced simply by the following code, which works for any geometry
```python
field = self.get_field( t=t, iteration=iteration, field='E',
 coord=pol, m=m, slicing_dir=slicing_dir)
```

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        extent = field[1]
+        if index == 'center':
+            # Get central slice
+            field_slice = field[0][int( field[0].shape[0] / 2), :]
+            # Calculate inverse FFT of filtered FFT array
+            envelope = self._fft_filter(field_slice, freq_filter)
+        elif index == 'all':
+            envelope = [ self._fft_filter(field[0][i, :], freq_filter)
+                         for i in range(field[0].shape[0]) ]
+        else:
```
RemiLehe: I would replace 
```python
 envelope = [ self._fft_filter(field[0][i, :], freq_filter)
 for i in range(field[0].shape[0]) ]
```
by 
```
envelope = np.array( [ self._fft_filter(field[0][i, :], freq_filter)
 for i in range(field[0].shape[0]) ])
```
Because it is generally nicer to return a 2d array than a list.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        -------
+        A 1D array with filtered input data in time/space domain
+        """
+        # Number of sample points
+        N = field.size
+        # Fourier transform of the field slice
+        fft_field_slice = np.fft.fft(field)
+        fft_freqs = np.fft.fftfreq(N)
+        # Find central frequency
+        central_freq_i = np.argmax(fft_field_slice[:N / 2])
```
RemiLehe: I am not sure about the following but wouldn't it make more physical sense to use the maximum of the absolute value
```python
central_freq_i = np.argmax( abs(fft_field_slice[:N / 2]) )
```
instead of the maximum of the real part
```python
central_freq_i = np.argmax( fft_field_slice[:N / 2] ) 
```
I guess I am a bit afraid that the modulations in the spectral phase could in principle lead the real part to be zero at the central frequency.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+                                                 pol=pol)[0])
+        # Get mean frequency
+        omega = self.get_mean_frequency(t=t, iteration=iteration, pol=pol)
+        # Calculate a0
+        a0 = Emax * const.e / (const.m_e * const.c * omega)
+        return( a0 )
+
+    def get_ctau( self, t=None, iteration=None, pol=None ):
+        """
+        Calculate the length of a (gaussian) laser pulse. ( sqrt(2) * sigma_z)
```
RemiLehe: Maybe the documentation could be slightly more precise here, e.g.
```
Calculate the length of a (gaussian) laser pulse.
(Here "length" means the "longitudinal waist", i.e. sqrt(2) times the RMS of the field)
```

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+        # Find the maximum of the envelope along the transverse axis
+        trans_max = np.amax(field, axis=1)
+        # Get transverse positons
+        trans_pos = getattr(extent, extent.axes[0])
+        # Calculate standard deviation
+        sigma_r = wstd(trans_pos, trans_max)
+        # Return the laser waist = sqrt(2) * sigma_r
+        return(np.sqrt(2) * sigma_r)
+
+    def wigner_transform( self, t=None, iteration=None, pol=None ):
```
RemiLehe: Nice ! Another really great feature !!!

I have a suggestion on the API: why not return a (properly initialized) `FieldMetaInformation` object along with the wigner transform. That way, the user can plot the wigner transform by using
`plt.imshow( wigner, extent=info.imshow_extent )`

See below for how the FieldMetaInformation could be initialized.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -17,4 +20,668 @@ def __init__(self, path_to_dir):
...
+                E_shift[:itau] = 0
+            EE = E * E_shift ** 2
+            fftwigner = np.fft.fft(EE)
+            wigner[i, :] = np.abs(fftwigner) ** 2
+        # Rotate and flip array to have input form of imshow
+        wigner = np.flipud(np.rot90(wigner[:, E.size / 2:]))
+        # Calculate the axis range
+        maxi, maxj = np.unravel_index(wigner.argmax(), wigner.shape)
+        xmin = -(T - T / wigner.shape[1] * maxj)
+        xmax = -(0 - T / wigner.shape[1] * maxj)
```
RemiLehe: At this point, instead of the above lines
```python
 maxi, maxj = np.unravel_index(wigner.argmax(), wigner.shape)
 xmin = -(T - T / wigner.shape[1] * maxj)
 xmax = -(0 - T / wigner.shape[1] * maxj)
```
I would initialize a FieldMetaInformation object:
```
info = FieldMetaInformation( {0:'t', 1:'omega'}, wigner.shape,
         gridspacing=(dt, np.pi/dt/wigner.shape), grid_unitSI=1.,
        global_offset=(0,0), position=(0,0))
return( wigner, info )
```
but the above is most certainly wrong: I am not sure my `gridspacing` is correct, and my `global_offset` (the start of the grid along each axis) is certainly wrong. Do you think you could adapt the above to make it right ?

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: I will check if that works. It took some time to wrap my head around this, but I agree this would be a faster option.

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: Yes!

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: I fully agree with you here. I did not think about the possibility of combining the two results. Great idea!

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: And it might also be a little bit faster to iterate over the numpy array.

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: I agree. Now that I notice this I am actually surprised that this works at all since a complex is passed to `np.argmax`.

soerenjalas:
Thanks for the great comments! I agree with you on all of them. I will make the changes in the next few days.

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: Okay, I did not think about that case. But this should be an easy fix

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: Yes, this is very elegant.

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: However, I'm not quite sure about giving `theta` and `slicing_dir` as optionals, since they are implicitly determined by `pol`. So we should probably agree on one option.

soerenjalas:
I implemented the changes according to your comments.
Here a few remarks:
* Using `np.histogram` in `get_current` apparently works and it's blazing fast compared to the old algorithm 
* by using 'x' and 'y' as arguments for `pol` arbitrary polarisations are not callable in methods that use `get_envelope`. In my eyes this should be uncritical in most use cases, but it is somewhat a restriction.
* The import of `FieldMetaInformation` looks rather nasty... Maybe we should but it into the init.py of opmd_viewer. Or we just live with it :)
* In `get_envelope` the extent of the `get_field` call is returned with obsolete r information in the 1D case. I can think of two solutions for this:
   * Construct a new extent object in the method
   * Give the  `FieldMetaInformation` object a method to delete single coordinates.

I don't know how much this will be needed so the second option might be a little overkill. Which do you prefer?

RemiLehe:
Great! Thanks for the update of the pull request. 

I'll merge this version with `dev`, and will answer the remaining points through pull requests of my own.

I think we can expect to merge this with master by the end of the week.

PR #11, opened 2015-10-30
Modifications of the LPA diagnotics

RemiLehe:
I did a couple of modifications on the LPA diagnostics:

- Simplify the import of FieldMetaInformation by bringing it at the `opmd_viewer` level

- Added a method of FieldMetaInformation to restrict the data to 1D

- Modified the `get_mean_frequency` function:
Instead of taking the RMS value of the frequency, I took the frequency for which the spectrum is maximal. This is because, in testing the function on simulation, I found that even a small amount of noise at high frequency could have a large impact on the RMS, and thus shift significantly the value from the expected one. Searching for the frequency of the maximum avoids this. However, I agree that taking the frequency of the maximum also has many problem, and thus I not entirely decided on that. It'd be good to have your opinion on this.

- Modified the `wigner_transform` function. I mainly changed the name of the function and the variables, because the operation that is performed is actually not - mathematically speaking - a wigner transform. I also modified some parts of the function in an attempt to simplify it a bit, but please feel free to comment on those change below if you don't agree or would like to see modifications.

Thanks,

### DIFF
--- a/opmd_viewer/__init__.py
+++ b/opmd_viewer/__init__.py
@@ -1,3 +1,3 @@
 # Make the OpenPMDTimeSeries object accessible from outside the package
-from .openpmd_timeseries import OpenPMDTimeSeries
-__all__ = ['OpenPMDTimeSeries']
+from .openpmd_timeseries import OpenPMDTimeSeries, FieldMetaInformation
+__all__ = ['OpenPMDTimeSeries', 'FieldMetaInformation']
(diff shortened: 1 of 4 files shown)


soerenjalas:
Thanks for the changes! I'm ok to merge them.

RemiLehe:
Thanks for reviewing these changes! This is now part of `master`.

PR #13, opened 2015-10-31
Laser spectrum function

RemiLehe:
This pull request simply introduces the method `get_spectrum` for the LpaDiagnostic (1D on-axis spectrum as a function of omega).

Not quite as fancy as the `get_spectrogram function`, but it could still be useful!


soerenjalas:
Nice, thanks! This could especially be useful to see if `get_main_frequency` makes sense in all cases.

PR #14, opened 2015-11-02
Fixed omega axis increment in get_spectrogram

soerenjalas:
I found a small bug in `get_spectrogram` : The omega increment should be 2*Pi/T


RemiLehe:
Good catch! Sorry for this mistake, and thanks for spotting it! 
I'll push this to `master` so that it also gets fixed.

PR #19, opened 2015-11-10
Adapt viewer to openPMD STANDARD (backward compatible)

RemiLehe:
The openPMD STANDARD will soon be publicly released. Therefore, I had another look at it and found that the viewer was overlooking some aspects:

- The possible existence of particlePatches: this commits explicitly tells
the algorithm to ignore the particlePatches
- The actual form of the basePath (the viewer assumed a string of the
from '/data/0/' whereas the standard specifies that the string should be of
the form '/data/%T')

This pull request corrects those problems. These changes are backward compatible, meaning that this will still work with current output of Warp and FBPIC.


RemiLehe:
I tried to merge that with `master` instead of `dev` and therefore I closed this pull request and created a new one.

PR #24, opened 2015-11-10
Make selection functional in LpaDiagnostics

RemiLehe:
As we discussed earlier, the particle diagnostics should be able to take advantage of the particle selection features. 
This pull request implements this. 
I tested the get_charge, get_emittance and get_divergence methods in this context.


soerenjalas:
Nice! I was literally about to start doing this :)

PR #25, opened 2015-11-10
Correct calculation of divergence

RemiLehe:
There was a small typo in the calculation of the divergence


soerenjalas:
Oh, stupid mistake! Thanks for fixing that!

RemiLehe:
Actually I am maybe the culprit for that bug. I think I committed it when I introduced the arctan functions. Sorry...

PR #26, opened 2015-11-10
Normalize the spectrum of the laser

RemiLehe:
This introduces a dt coefficient in the calculation of the spectrum
so that the spectrum really corresponds to the discretization of
tmpremi

Without the dt coefficient, the amplitude of the spectrum scales
linearly with the number of gridpoints along z, which is unwanted when comparing
spectra for two different resolution.


ax3l:
@RemiLehe if you want you can use [LaTeXiT](http://pierre.chachatelier.fr/latexit/latexit-home.php?lang=en) for mac (or `klatexformula` for Linux) and attach the files via drag&drop in the description :)

PR #29, opened 2015-12-03
Fixed an erroneous warning message

RemiLehe:
There was a minor mistake in the message about the available iterations.


soerenjalas:
Oh, I didn't notice this even though I probably came across it several times... Thanks for fixing!

PR #32, opened 2015-12-30
Add executable `openPMD_notebook`

RemiLehe:
For new users, it is not easy to remember what to write inside a new IPython notebook, in order to have the GUI working. 

Therefore, I included a small executable script that creates a pre-filled notebook and automatically opens it in a browser. Please see the README.md for documentation on that script.


soerenjalas:
This is really nice. Thanks!

PR #33, opened 2015-12-31
Close #22 : Add Continuous Integration

RemiLehe:
**Please merge pull request #32 first**

## Overview

Following the advice of @ax3l, this pull requests:
- Creates a set of automated tests. These tests can now be run by typing `python setup.py test`, as mentioned in CONTRIBUTING.md.
- Links them to [Travis CI](https://travis-ci.org/), so that these tests are automatically run every time someone pushes commits to the Github repository. (The tests run on remote servers that provided for free by Travis CI.)

I heavily relied on [Open Sourcing a Python Project the Right Way](https://www.jeffknupp.com/blog/2013/08/16/open-sourcing-a-python-project-the-right-way/), when doing these changes.

NB: I also updated some of the installation instructions as they were no longer valid.

## More details on the automated tests

For the moment they are quite basic: they simply make sure that the notebooks in the `tutorials` directory run without any problem. This is done by converting the notebooks to pure-Python scripts and running these scripts. Therefore, the features that are specific to the IPython notebook (mainly the GUI) are not tested (I did not find a way to test them, but it would be nice to have one !).

## More details on the tests on Travis CI

The Travis CI instructions are given in `.travis.yml`. The tests include:
- Installing openPMD-viewer from scratch, by first installing a light version of Anaconda ([Miniconda](http://conda.pydata.org/miniconda.html)), installing all the listed dependencies of openPMD-viewer, and installing openPMD-viewer itself.
- Running `python setup.py test` to check the tutorial notebooks.

This is done for each push and also each pull requests, resulting in e.g. a nice button below saying that "All checks have passed".

I have set up Travis CI for my own fork of openPMD-viewer, and for the main repository. But I think that for the other forks, their owner have to register to Travis CI themselves if they want the tests to be run on their forks. See [Getting started - Travis CI](https://docs.travis-ci.com/user/getting-started/)

PR #34, opened 2016-01-04
Python 3.X Support

ax3l:
Follow up to #33: Adds a Python 3.X test to CI via travis. ~~Currently, due to errors in it we allow a 3.X failure without marking the PR as invalid. (Nevertheless, we might want to add python 3 support.)~~

At least for what is covered by the tests, this PR adds Python 3.X support to openPMD-viewer (alongside its Python 2 support).

Additionally, some missing end-of-file empty newlines are added for better diff readability.


tests/test_tutorials.py
```diff
@@ -31,7 +34,13 @@ def test_tutorials():
         script_name = notebook_name[:-6] + '.py'
         os.system( 'ipython nbconvert --to=python %s' %notebook_name )
         clean_ipython_features( script_name )
-        execfile( script_name )
+        try:
+            exec( open(script_name).read() )
```
ax3l: Note: I am still *sourcing* the file which is probably not ideal and can have side-effects.

Using `os.system()` and checking its return status is probably the better choice for a follow-up refactoring.

tests/test_tutorials.py
```diff
@@ -54,12 +63,12 @@ def clean_ipython_features( script_name ):
 
         # Replace the lines that activate matplotlib in a notebook
         # by a line that selects the PS backend
-        if re.match("[ ]*get_ipython\(\)\.magic\(u'matplotlib",
+        if re.match("[ ]*get_ipython\(\)\.magic\((.*)'matplotlib",
```
ax3l: utf8 just became standard in python3 so we expect *zero or more* chars here

RemiLehe:
Thanks for the code improvements ! And thanks a lot for debugging the Python 3 version !!

I have a few minor changes, which I'll submit via a PR to this branch.

ax3l:
sweet, thanks I will wait for them! :)

RemiLehe:
Here it is : 
https://github.com/ax3l/openPMD-viewer/pull/1

ax3l:
merged and updated (I re-added the "print script on error" functionality)

RemiLehe:
Great ! Thanks for the modifications.

PR #36, opened 2016-01-06
Added pyflakes test

RemiLehe:
This adds pyflakes testing, similarly to what @ax3l added to the openPMD-validator.


ax3l:
pyflakes likes that :+1:

PR #41, opened 2016-01-19
LpaDiagnostics tutorial

soerenjalas:
Hi, 
I wrote a tutorial notebook (highly based on the ones you made) for the Lpa addon class, with the main idea to have at least some degree of automatic testing done for the methods. I'm not quite sure yet if this is a good way to do this, because for one it does not really cover all cases that could be tested (only 2D at the moment), and then I'm not sure how helpful it is as a tutorial itself.
Maybe you could tell you thoughts on this or if a dedicated testing script is maybe the better choice. 

The next thing I wasn't sure about was where to put this in the repo. Right now it is in the same directory as the other tutorials, but maybe we should separate the 'addon documentation' from the main stuff.

So please be open about your thoughts on this... will not take it personally :)

### DIFF
--- a/opmd_viewer/addons/__init__.py
+++ b/opmd_viewer/addons/__init__.py
@@ -1,0 +1,2 @@
+from .pic import LpaDiagnostics
+__all__ = ['LpaDiagnostics']
--- a/opmd_viewer/addons/pic/__init__.py
+++ b/opmd_viewer/addons/pic/__init__.py
@@ -1,0 +1,2 @@
+from .lpa_diagnostics import LpaDiagnostics
+__all__ = ['LpaDiagnostics']
--- a/tutorials/4_LPA-tools.ipynb
+++ b/tutorials/4_LPA-tools.ipynb
@@ -517,15 +517,6 @@
    "source": [
     "ts_2d.get_ctau(iteration=200, pol='y')"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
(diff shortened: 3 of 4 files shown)


RemiLehe:
Hi Soeren,

Thanks for adding the tutorial (and sorry for my late reply). 
I think that it is good as a tutorial notebook, as it exposes many of the functions that users may not know about. (Most of these functions don't need much explanation ; but users have to know that they exist.)

Also, I don't have yet a definite opinion on this, but I would say that separate test scripts could be most useful when testing the **validity** of the results (e.g. is the calculation of a0 correct) which is not the case here. To me, a notebook is fine.

One thing I am wondering is if we should not maybe simplify the import for LpaDiagnostics : for instance `from opmd_viewer.addons import LpaDiagnostics` instead of `from opmd_viewer.addons.pic.lpa_diagnostic import LpaDiagnostics` (this can be done by modifying the __init__.py files). @ax3l : Any opinion on this ?

In the meanwhile, I'll do a few minor modifications directly on the pull request.

ax3l:
@RemiLehe sorry for answering a bit late, I was close to finish my answer before leaving today when you merged :)

I think the short cut import is useful and convenient, although I must admit that a linear lpa can be also modelled without PIC in some cases ;)

I think it would be useful if we could further auto-test the notebook if it is still using the correct syntax (similar as we do in tests), otherwise it will get outdated rather quickly I guess.

PR #46, opened 2016-02-29
Update README.md

RemiLehe:


README.md
```diff
@@ -13,6 +13,9 @@ contents of a set of [openPMD](http://www.openpmd.org/#/start) files
 To install this package :
 
 - Clone this repository using `git`
+```
+git clone https://github.com/openPMD/openPMD-viewer.git
```
ax3l: the anonymous, read-only link without a github account would be
```bash
git clone git://github.com/openPMD/openPMD-viewer.git
```

RemiLehe:
Are you sure that `git clone https://github.com/openPMD/openPMD-viewer.git` does not work ? It is the adress that is given on the github page of openPMD when you access it without signing in. 

Also, I tried it within a docker container (which should be anonymous, in the sense that it has no knowledge of my name or the fact that I have a Github account), and it seemed to work.

ax3l:
Cool, I sometimes have problems with `https` and anonymous clones but if it works for you then we can keep it.

PR #52, opened 2016-04-12
LPA utils: Slice gamma spread method

soerenjalas:
After doing some work with beams for the last couple of weeks I added a method I was using quite often to the LPA utilities. Even though this is a quite simple approach by looping over the slices, the method performs quite OK for reasonable slice widths.


RemiLehe:
Great! Thanks for adding this method.

PR #57, opened 2016-04-20
openPMD Extensions: Implement Detection

ax3l:
This pull request implements the correct named detection of [openPMD extensions](https://github.com/openPMD/openPMD-standard/blob/1.0.0/STANDARD.md#hierarchy-of-the-data-file).

Since a file can implement several extensions at the same time, the extension is encoded as a [bitmask](https://en.wikipedia.org/wiki/Mask_%28computing%29). The user does not care about the bit mask but about a final list of extensions that are supported :)

### API Changes

- `OpenPMDTimeSeries.extension`
  - removed
  - (did contain the [bitmask](https://en.wikipedia.org/wiki/Mask_%28computing%29))
  - (not used in the code yet nor in examples)
- `OpenPMDTimeSeries.extensions`
  - added (plural!)
  - contains a list of named extensions
  - detection and name matching is done in `params_reader.py`
    for each file (same implementation level as version)

### Usage

A user can now query
```python
from opmd_viewer import OpenPMDTimeSeries

ts = OpenPMDTimeSeries('...')
print(ts.extensions)
# []
# ['ED-PIC']
# ['ED-PIC', 'An-other-fancy-extension']

if 'ED-PIC' in ts.extensions:
    print('I will now do PIC specific stuff!')
```

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
@@ -5,6 +5,7 @@ It defines a function that can read standard parameters from an openPMD file.
 """
 import os
 import h5py
+import numpy as np
 from .utilities import is_scalar_record, get_shape, get_bpath
 
 def read_openPMD_params( filename ):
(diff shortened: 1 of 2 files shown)


opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
```diff
@@ -31,7 +32,15 @@ def read_openPMD_params( filename ):
...
-    params['extension'] = f.attrs['openPMDextension']
+
+    # Find out supported openPMD extensions claimed by this file
+    # note: a file might implement multiple extensions
+    known_extensions = {'ED-PIC': np.uint32(1)}
+    bitmask_all_extensions = f.attrs['openPMDextension']
+    params['extensions'] = []
+    for extension, bitmask in known_extensions.items():
+        if bitmask_all_extensions & bitmask == bitmask:
+            params['extensions'].append(extension)
```
RemiLehe: Put a link to the standard and link to wikipedia bitmask

PR #58, opened 2016-04-20
Record Component Detection: Per Species

ax3l:
It is a too strict assumption that all species must be identical in their records. This commit adds a `species_record_components` attribute created by `read_openPMD_params` that is a dictionary for each species containing the usual list of record components.

I unified the naming a bit since it is useful if we stay with the naming in the standard ("record component") and not introduce new namings (such as "particle quantity").

### Previous Implementation

Since the old attribute `avail_ptcl_quantities` is
- assuming all species are identical
- using an undefined naming

let us internally "deprecate" it for removal in the future. (Nevertheless, for now it still works as before.) Unfortunately, I found no easy way to throw a warning on usage since one would need to wrap the `__getattr[ibute]__` member for that with a base class.

### Change in API

In case no species are found, the old member `avail_ptcl_quantities` (and the new member `species_record_components`) are still *defined* and set to `None` instead of letting them undefined. The simple reason for that is, that catching a "member not found" is more expensive and more verbose then simply iterating over a `None` (empty) but *existing* member.

### Further Work

For the `interactive.py` class we should add an event handling that is refreshing the `record components` of particles as soon as a user selected a new `species`.

This can be done in an independent PR into `dev`.

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
@@ -82,29 +82,42 @@ def read_openPMD_params( filename ):
         params['avail_species'] = []
         for species_name in bpath[particle_path].keys():
             params['avail_species'].append(species_name)
-        # Extract the available particle quantity, from the first species
-        first_species_path = next(iter(params['avail_species']))
-        first_species = bpath[os.path.join(particle_path, first_species_path)]
-        ptcl_quantities = []
-        # Go through all the particle quantities
-        for quantity_name in first_species.keys():
-            # Skip the particlePatches, which are not used here.
-            if quantity_name == 'particlePatches':
-                continue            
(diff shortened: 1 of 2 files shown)


opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
```diff
@@ -82,76 +82,89 @@ def read_openPMD_params( filename ):
...
+                else:
+                    # Add each component of the vector record
+                    for coord in record.keys():
+                        species_record_components[species_name]. \
+                            append(os.path.join(record_name, coord))
+            # Simplify the name of some standard openPMD records
+            species_record_components[species_name] = \
+                simplify_record( species_record_components[species_name] )
+        params['avail_species_record_components'] = species_record_components
+        # deprecated
```
RemiLehe: Please remove the deprecated attributes completely

PR #59, opened 2016-04-20
Updated the notebook starter

RemiLehe:
The command 'ipython notebook' is now deprecated, and is replaced by 'jupyter notebook'


ax3l:
the push is confused since the branch was first pushed to `mainline` and is now half-cached.
builds still fine.

PR #62, opened 2016-04-20
Allow species with different records in GUI

RemiLehe:
- updates the record components in the `slider()` GUI every time a new species is selected

### DIFF
--- a/opmd_viewer/openpmd_timeseries/interactive.py
+++ b/opmd_viewer/openpmd_timeseries/interactive.py
@@ -33,7 +33,6 @@ class InteractiveViewer(object):
         kw: dict
             Extra arguments to pass to matplotlib's imshow
         """
-
         # -----------------------
         # Define useful functions
         # -----------------------
@@ -71,7 +70,7 @@ class InteractiveViewer(object):
                     slicing=slicing_button.value, theta=theta_button.value,
                     slicing_dir=slicing_dir_button.value,
                     vmin=vmin, vmax=vmax, cmap=fld_color_button.value )
-                
+
         def refresh_ptcl(force=False) :
             "Refresh the current particle figure"
             
@@ -119,8 +118,14 @@ class InteractiveViewer(object):
 
         def refresh_species(b):
             """
-            Refresh the particle species buttons
+            Refresh the particle species buttons by populating them
+            with the available records for the current species
             """
+            # Deactivate the particle refreshing to avoid callback
+            # while modifying the widgets
+            saved_refresh_value = ptcl_refresh_toggle.value
+            ptcl_refresh_toggle.value = False
+            
             # Get available records for this species (remove charge and
             # mass as they are typically less interesting)
             avail_records = [ q for q in \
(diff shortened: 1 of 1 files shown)


opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -116,7 +116,27 @@ def refresh_ptcl(force=False) :
...
+
+        def refresh_species(b):
+            """
+            Refresh the particle species buttons
+            """
+            # Get available records for this species (remove charge and
+            # mass as they are typically less interesting)
+            avail_records = [ q for q in \
+                        self.avail_record_components[ptcl_species_button.value]
+                        if (q in ['charge', 'mass'])==False ]
```
ax3l: let us make this an `exclude_list=...` option of `slider()` in the future so users can choose what to exclude

opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -116,7 +116,27 @@ def refresh_ptcl(force=False) :
...
+            ptcl_xaxis_button.value = avail_records[0]
+            ptcl_yaxis_button.options = avail_records + ['None']
+            ptcl_yaxis_button.value = 'None'
+            # Update the selection widgets
+            for dropdown_button in ptcl_select_widget.quantity:
+                dropdown_button.options = avail_records
+            
+            refresh_ptcl()
+                    
         def refresh_ptcl_now(b):
```
ax3l: note for future improvements: you can use [`functools.partial`](https://docs.python.org/2/library/functools.html#functools.partial) for that job: it binds default arguments to functions so you can use them in callbacks with the right signature

```python
from functools import partial

refresh_ptcl_now = partial(refresh_ptcl, True)
ptcl_refresh_button.on_click( refresh_ptcl_now )
```

(C++ term: binding)

opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -116,7 +116,27 @@ def refresh_ptcl(force=False) :
...
+            Refresh the particle species buttons
+            """
+            # Get available records for this species (remove charge and
+            # mass as they are typically less interesting)
+            avail_records = [ q for q in \
+                        self.avail_record_components[ptcl_species_button.value]
+                        if (q in ['charge', 'mass'])==False ]
+            # Update the plotting buttons
+            ptcl_xaxis_button.options = avail_records
+            ptcl_xaxis_button.value = avail_records[0]
```
ax3l: I think this should read
```python
ptcl_xaxis_button.value = avail_records[ptcl_species_button.value]
```

no?

opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -274,24 +294,26 @@ def step_bw(b) :
...
-            avail_ptcl_quantities = [ q for q in self.avail_ptcl_quantities \
+            ptcl_species_button.on_trait_change( refresh_species )
+            # Get available records for this species (remove charge and
+            # mass as they are typically less interesting)
+            avail_records = [ q for q in \
+                        self.avail_record_components[ptcl_species_button.value]
                         if (q in ['charge', 'mass'])==False ]
             # Particle quantity on the x axis
             ptcl_xaxis_button = widgets.ToggleButtons(
-                value='z', options=avail_ptcl_quantities )
```
ax3l: maybe we can create the buttons with an empty list here and fill them with a `refresh` at the end.

that would reduce code duplication and separate layouting from data filling.

ax3l:
very cool, thank you! :sparkles:

PR #63, opened 2016-04-20
`get_particle`: Speed Up Read

ax3l:
This commits speeds up reads in `get_particle` by keeping the file opened between the calls to `get_species_data`.

File opening-closing is an operation with high latencies and should therefore be minimized.

The internal `get_particle` interface has been renamed to openPMD namings. Instead of the `filename` a `file_handle` with an already opened (*read*) `h5py.File` object is passed.


RemiLehe:
Nice PR!

PR #65, opened 2016-04-20
Update `nbconvert`: use `jupyter`

ax3l:
Same update as in #59 to get rid of the `ipython notebook` is *deprecated* warning.


RemiLehe:
Nice !

PR #67, opened 2016-04-20
Close #54: Skip Check of other Files 

### LINKED ISSUE #54: OpenPMDTimeSeries: Skip Check of other Files
To save time while opening an `OpenPMDTimeSeries` we could add an optional parameter that is only querying the first file for, e.g., available particle records and assumes the other files will not change the available attributes for each species over time.

RemiLehe:
Added a flag that prevents checking of all files (this can be very slow sometimes)


ax3l:
thank you, great! :)

PR #71, opened 2016-04-21
Handle macroweighting

RemiLehe:
This pull request implements correct reading of macroweighted quantities. 

In order to test it thoroughly, it would be great if we could have a PIConGPU output with the macroWeighted flag set properly.


ax3l:
@RemiLehe we need to rebase this branch against `dev` due to the formatting updates in the PEP8 styling.

ax3l:
merci! `unitSI` might be missing, need to dbl check

PR #74, opened 2016-04-21
Fix the GUI

RemiLehe:
The interactive GUI used to crash before this PR.
The PR also removes the deprecated `on_trait_change` and replaces
it with the proper command `observe`:
https://ipywidgets.readthedocs.org/en/latest/examples/Widget%20Events.html

### DIFF
--- a/opmd_viewer/openpmd_timeseries/interactive.py
+++ b/opmd_viewer/openpmd_timeseries/interactive.py
@@ -352,7 +352,8 @@ class InteractiveViewer(object):
             # Particle selection
             # ------------------
             # 3 selection rules at maximum
-            ptcl_select_widget = ParticleSelectWidget(3, refresh_ptcl)
+            ptcl_select_widget = ParticleSelectWidget(3,
+                                 avail_records, refresh_ptcl)
 
             # Plotting options
             # ----------------
@@ -446,7 +447,7 @@ class ParticleSelectWidget(object):
     Class that groups the particle selection widgets.
     """
 
-    def __init__(self, n_rules, refresh_ptcl):
+    def __init__(self, n_rules, avail_records, refresh_ptcl):
         """
         Initialize a set of particle selection widgets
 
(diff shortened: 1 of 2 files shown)


opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -39,21 +39,20 @@ def slider(self, figsize=(10, 10), **kw):
         # Define useful functions
         # -----------------------
 
-        def refresh_field(sender_change=None, force=False):
+        def refresh_field(change, force=False):
```
ax3l: let us keep `change=None` so we can still call it without the object that we do not use anyway

opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -87,23 +86,21 @@ def refresh_field(sender_change=None, force=False):
                     slicing_dir=slicing_dir_button.value,
                     vmin=vmin, vmax=vmax, cmap=fld_color_button.value)
 
-        def refresh_ptcl(sender_change=None, force=False):
+        def refresh_ptcl(change, force=False):
```
ax3l: same here, let us keep `=None`

opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -146,10 +143,17 @@ def refresh_ptcl(sender_change=None, force=False):
                         vmin=vmin, vmax=vmax, cmap=ptcl_color_button.value,
                         nbins=ptcl_bins_button.value)
 
-        def refresh_species(b):
+        def refresh_species(change):
```
ax3l: same here

opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -173,13 +177,13 @@ def refresh_species(b):
...
+            ptcl_refresh_toggle.value = saved_refresh_value
 
-        def change_t(name, value):
+        def change_t(change):
             "Plot the result at the required time"
-            self.current_t = 1.e-15 * value
-            refresh_field()
-            refresh_ptcl()
+            self.current_t = 1.e-15 * change['new']
+            refresh_field( change )
```
ax3l: you could keep `()` because forwarding the caller object might not help at this point (it is already handled)

ax3l:
fantastic, thank you! :sparkles:

PR #79, opened 2016-04-22
Deactivate the coordinate widget for scalar fields

RemiLehe:
it makes no sense to give the user the possibility to query 'x', 'y', 'z' when the field is for instance the charge density. 

This pull request deactivates the corresponding widget for scalar fields.


ax3l:
very nice, thanks!

PR #80, opened 2016-04-22
Close #56 License Headers

### LINKED ISSUE #56: License Headers
We should add a license header to each file stating the 3-Clause-BSD-LBNL license is used.
This is good practice since derivatives might only clone a specific file.

I can prepare a snippet later on, not urgent.

```python
"""
File description 
...

__authors__ = "Remi Lehe, Soeren Jalas, Axel Huebl, ..." (from `git log` of each file)
__copyright__ = "Copyright 2015-2016"
__credits__ = ["Remi Lehe", "Soeren Jalas", "Axel Huebl"]
__license__ = "3-Clause-BSD-LBNL"
"""
```

([addLicense](https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.1.2/src/tools/bin/addLicense) tool in PIConGPU)

Also: avoid UTF-8 for now, write names in pure ascii (sometimes crashes people's setups otherwise)

ax3l:
Close #56 by adding a short license and author information to each file (as an open source best-practice, since files might be individually re-used in derivative projects).

The authors were taken from the individual `git log` history of each file, new authors can update the author list of edited files as they contribute.

Unified some spellings of the project in the way (lowercase "o" in "openPMD", most often written without "-" between "openPMD" and "viewer").

### DIFF
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -1,3 +1,13 @@
+"""
+This file is part of the openPMD viewer.
+
+This file contains diagnostics relevant for laser-plasma acceleration
+
+__authors__ = "Soeren Jalas, Remi Lehe"
+__copyright__ = "Copyright 2015-2016, openPMD viewer contributors"
+__license__ = "3-Clause-BSD-LBNL"
+"""
+
 # Class that inherits from OpenPMDTimeSeries, and implements
 # some standard diagnostics (emittance, etc.)
 from opmd_viewer import OpenPMDTimeSeries, FieldMetaInformation
--- a/opmd_viewer/openpmd_timeseries/data_reader/field_metainfo.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/field_metainfo.py
@@ -1,10 +1,15 @@
 """
-This file is part of the OpenPMD viewer.
+This file is part of the openPMD viewer.
 
 It defines the main FieldMetaInformation class, which
 is returned by `get_field` along with the array of field values,
 and gathers information collected from the openPMD file.
+
+__authors__ = "Remi Lehe"
+__copyright__ = "Copyright 2015-2016, openPMD viewer contributors"
+__license__ = "3-Clause-BSD-LBNL"
 """
+
 import numpy as np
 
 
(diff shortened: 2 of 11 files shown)


opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -1,10 +1,19 @@
+"""
```
RemiLehe: We usually start with "This file is part of the openPMD viewer."

For instance here:
```
"""
This file is part of the openPMD viewer.

It contains diagnostics relevant for laser-plasma acceleration.
```

RemiLehe:
I would have a slight preference for adding the __author__, __license__, etc. just **before** the import statements. (That way we have first the docstring for the file and the __author__, __license__, etc.`, and **then** the actual Python code, starting with the import statements)

What is your opinion on this ?

opmd_viewer/addons/pic/lpa_diagnostics.py
ax3l: oh yes, that should be in! will update!

ax3l:
> I would have a slight preference for adding the author, license, etc. just before the import statements. 

Yes, I initially did this since it is way more obvious! But PEP8 did not allow that, it expects imports to come first... :sa: 

Otherwise we could maybe add it in the docstring itself?

```python
This file is part of the openPMD viewer.

It contains diagnostics relevant for laser-plasma acceleration.

__authors__ = "Soeren Jalas, Remi Lehe"
__copyright__ = "Copyright 2015-2016, openPMD viewer contributors"
__license__ = "3-Clause-BSD-LBNL"
"""```

**Comment by RemiLehe** [comment]
How about we do something similar to https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/datasets/base.py. 
In our case, for instance, this could give:
```python
This file is part of the openPMD viewer.

It contains diagnostics relevant for laser-plasma acceleration
"""
# Copyright 2015-2016 Soeren Jalas, Remi Lehe
# License: 3-Clause-BSD-LBNL

 from opmd_viewer import OpenPMDTimeSeries, FieldMetaInformation
 import matplotlib.pyplot as plt
 import numpy as np
 import scipy.constants as const
```

If you agree with the above, I can update the pull request.

ax3l:
that is kind of the non-pythonic way.

I would go with the one I proposed by putting it in the file's docstring and update it. is that ok for you?

ax3l:
Can also do it like this:
```python
"""
This file is part of the openPMD viewer.

It contains diagnostics relevant for laser-plasma acceleration.

Copyright 2015-2016 Soeren Jalas, Remi Lehe
License: 3-Clause-BSD-LBNL
"""

import numpy as np
# ...
```

RemiLehe:
Ok for the last one. I'll update the PR.

ax3l:
Will do, sure! :)

PR #81, opened 2016-04-22
Multiply by unitSI only if different than 1

RemiLehe:
This makes reading data more efficient.

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/utilities.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/utilities.py
@@ -89,7 +89,8 @@ def get_data(dset, i_slice=None, pos_slice=None):
             data = dset[:, :, i_slice]
 
     # Scale by the conversion factor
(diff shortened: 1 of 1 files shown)


ax3l:
hehe, like it :rocket:

opmd_viewer/openpmd_timeseries/data_reader/utilities.py
```diff
@@ -89,7 +89,8 @@ def get_data(dset, i_slice=None, pos_slice=None):
             data = dset[:, :, i_slice]
 
     # Scale by the conversion factor
-    data = data * dset.attrs['unitSI']
+    if dset.attrs['unitSI'] != 1:
```
ax3l: 1.0 would be a tiny bit better to avoid he cast from int to float

ax3l: will update it.

PR #83, opened 2016-04-22
Close #66 Slider: Record Exclude List 

### LINKED ISSUE #66: Slider: Record Exclude List
The `slider()` could add an optional `exclude_particle_records=list` argument which can be used to exclude `charge` and `mass` as its currently hard-coded.

RemiLehe:
Close #66 This gives the user the option to exclude certain particle quantities from the particle GUI


ax3l:
thank you, that was quick :)

PR #84, opened 2016-04-22
Replace nbins slider by a text widget

RemiLehe:
One of our users requested that the bins widget be a text widget instead of a slider. I actually agree with him, since changing the number of bins via a text widget is much more flexible.


ax3l:
great idea, thanks for the idea and the update!

PR #88, opened 2016-04-23
Add the versioning number

RemiLehe:
This is done in a similar way as https://jeffknupp.com/blog/2013/08/16/open-sourcing-a-python-project-the-right-way/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 from setuptools import setup, find_packages
+import opmd_viewer # In order to extract the version number
 
 # Get the package requirements from the requirements.txt file
 with open('requirements.txt') as f:
@@ -9,7 +10,7 @@ install_requires.append('wget')
 
 # Main setup command
 setup(name='opmd_viewer',
-      version='1.0',
+      version=opmd_viewer.__version__,
       maintainer='Remi Lehe',
       maintainer_email='remi.lehe@lbl.gov',
       description='Visualization tools for OpenPMD files',
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 from setuptools import setup, find_packages
-import opmd_viewer # In order to extract the version number
+import opmd_viewer  # In order to extract the version number
 
 # Get the package requirements from the requirements.txt file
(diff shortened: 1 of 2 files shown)


ax3l:
haha, I know that blog post, it's a good one! :sparkles:

setup.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 from setuptools import setup, find_packages
+import opmd_viewer  # In order to extract the version number
```
ax3l: this is part I am not 100% sure in the original blog post with:

the include has quite some dependencies. we could also create a single version.py file that we both include in `setup.py` and in ` o[pen]pmd_viewer/__init__.py`. By that we could depend on a more lightweight include with fewer side effects and still have the version number in one place for both the packaging and the module.

RemiLehe:
Ok, how about adding the version in `opmd_viewer/version.py` instead of `opmd_viewer/version/version.py`? Any preference?

RemiLehe:
Actually, now that I think about it, I am not sure that the solution you proposed will work: as I understand it, your solution is to replace, in `setup.py`
```
from opmd_viewer import __version__
```
by
```
from opmd_viewer.version.version import __version___
```
However, when executing the above import statement, Python will start by executing the `opmd_viewer.__init__.py`, and thus it will still import all the packages that are associated with opmd_viewer.

Am I correct?

ax3l:
What I thought is that python will only execute `opmd_viewer.version.__init__.py` but not the overlaying `opmd_viewer.__init__.py` but you might be right. In that case just leave it as it is :)

RemiLehe:
I did a few tests and it confirms that the `opmd_viewer.__init__.py` get executed, even if you are only importing and sub-sub-module.

Should the current PR be merged then?

ax3l:
hm, then we leave it as it is I guess. thank you a lot for checking!

setup.py
```diff
@@ -9,7 +10,7 @@
 
 # Main setup command
 setup(name='opmd_viewer',
-      version='1.0',
```
ax3l: hopefully this does not cause problems if someone has already installed the current version and then goes to install the first release aka `0.1.0`

PR #89, opened 2016-04-23
Change formatting of the copyright

RemiLehe:
It does not really make sense to use the Python synthax (like `__license__ = "3-Clause-BSD-LBNL"`) inside a comment, since this does not define the variable `__license__` anyway.

Therefore, I rewrote these part in more human-friendly way. This is only a suggestion.


ax3l:
I think that's a good idea :)

PR #91, opened 2016-04-24
Drop support for IPython.html

RemiLehe:
Since we started using `observe` instead of the deprecated `on_trait_change` in `interactive.py`, we effectively stopped supporting the deprecated widget package `IPython.html`, and started only supporting `ipywidgets` (which is the official widget package for Python).

This PR makes it more explicit, by raising a message (see main.py) if `ipywidgets` is not installed.


ax3l:
merci!

ax3l:
@RemiLehe I am wondering if we might want to document "optional" or ["enhancing" packages](http://stackoverflow.com/questions/3664478/optional-dependencies-in-a-pip-requirements-file/3664685#3664685) in an `optional-requirements.txt` file to have them all together.

PR #93, opened 2016-04-25
Added CHANGELOG.md

RemiLehe:
This is a first draft, and will be updated as we add more stuff for 0.1.0 release.


ax3l:
cool looks great! let us write consequently `openPMD viewer` as we did in the README and headers (sorry for that, will update it directly if you like)

PR #94, opened 2016-04-28
Changed name of the software to openPMD-viewer

RemiLehe:


ax3l:
excellent, back to the roots :)

PR #97, opened 2016-05-02
Added ´check_all_files´ flag to LPA diagnostic class init

### LINKED ISSUE #98: `check_all_files=False` breaks slider and time access of data
Hi,
in its current implementation `check_all_files=False`, prevents the usage of the slider widget as well as the ability to call `get_field(t=50years...)`. This happens because `self.t` is acquired in the same for loop where the data checking is done:

```python

self.t = np.zeros( N_files )
...
if check_all_files:
            for k in range(1, N_files):
                t, params = read_openPMD_params(self.h5_files[k])
                self.t[k] = t
                for key in params0.keys():
                    if params != params0:
                        print("Warning: File %s has different openPMD "
                              "parameters than the rest of the time series."
                              % self.h5_files[k])
```

This results in a `self.t` array filled with zeros, if `check_all_files=False`.

Sorry for being not too familiar with the standard, but if the timestep is saved within the openPMD file, we could use that in order to determine the time from the iteration data. Otherwise ( if the timestep is not allowed to change, again I need to get more familiar with this stuff) we could only open two files and get the timestep from the difference in time and iteration.

EDIT:
"READ THE FREAKING MANUAL!!"

Turns out this is in fact stated in the standard, but not to the benefit of the problem:
```
dt
type: (float)
description: The latest time step (that was used to reach this iteration). This is needed at the iteration level, since the time step may vary from iteration to iteration in certain codes. 
```

Okay then I see no other way then opening all files to acquire the times.

Technically the api should work without the times so if we want to have a "fast" mode we could have a custom exception thrown whenever a user wants to use the fast mode and call a time argument.  

EDIT 2:
This could be done for example in ` _find_output(self, t, iteration)`

```python
def _find_output(self, t, iteration):
...
        # If a time is requested
        elif (t is not None):
            # NEW
            if self.t == None:
                raise APIError:
                   'Fast-read mode does not support time data acquisition'
            
           # Make sur the time requested does not exceed the allowed bounds
            if t < self.tmin:
                self.current_i = 0
            elif t > self.tmax:
                self.current_i = len(self.t) - 1
...
```

To conclude this, I think that this flag is very nice since checking large datasets can take a great amount of time. Therefore we should try to find an elegant solution to this problem. I also tried to use multiprocessing for the initialisation step, but my attempts failed with the really not so nice implementation of the python multiprocessing module... What are your thoughts on this?

soerenjalas:
Resolve #98 first.

Hi,
As the OpenPMDTimeSeries init() method is called when creating an instance of LpaDiagnostics, the flag `check_all_files` needed to be added there as well.

### DIFF
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -31,7 +31,7 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
             one file per iteration, and the name of the files should end
             with the iteration number, followed by '.h5' (e.g. data0005000.h5)
         """
-        OpenPMDTimeSeries.__init__( self, path_to_dir, 
+        OpenPMDTimeSeries.__init__( self, path_to_dir,
                                     check_all_files=check_all_files )
 
     def get_mean_gamma( self, t=None, iteration=None, species=None,
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -18,7 +18,7 @@ import scipy.constants as const
 
 class LpaDiagnostics( OpenPMDTimeSeries ):
 
-    def __init__( self, path_to_dir ):
+    def __init__( self, path_to_dir, check_all_files=True ):
         """
         Initialize an OpenPMD time series with various methods to diagnose the
         data
@@ -31,7 +31,8 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
             one file per iteration, and the name of the files should end
             with the iteration number, followed by '.h5' (e.g. data0005000.h5)
         """
-        OpenPMDTimeSeries.__init__( self, path_to_dir )
+        OpenPMDTimeSeries.__init__( self, path_to_dir, 
+                                    check_all_files=check_all_files )
 
     def get_mean_gamma( self, t=None, iteration=None, species=None,
                         select=None ):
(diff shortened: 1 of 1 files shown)


opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -31,7 +31,8 @@ def __init__( self, path_to_dir ):
             one file per iteration, and the name of the files should end
             with the iteration number, followed by '.h5' (e.g. data0005000.h5)
         """
-        OpenPMDTimeSeries.__init__( self, path_to_dir )
+        OpenPMDTimeSeries.__init__( self, path_to_dir,
```
ax3l: pls remove the trailing whitespace :)

ax3l:
thank you for the fix, we did not test the LPA diag, d'oh!

opmd_viewer/addons/pic/lpa_diagnostics.py
soerenjalas: PEP, always a pain... Will do :)

ax3l: true ;)

but EOL white spaces are not pretty in git and diff anyway ;)

RemiLehe:
Thanks for the pull request. 
Could you please also copy the definition of `check_all_files` in the docstring of `LpaDiagnostic` ?

ax3l:
fantastic, thank you! :)

soerenjalas:
No more lazy afternoon coding, with all this official 1.0.0 stuff going on :)

ax3l:
yeaha :)

PR #100, opened 2016-05-03
Close #98 : `check_all_files=False` breaks slider and time access of data

### LINKED ISSUE #98: `check_all_files=False` breaks slider and time access of data

RemiLehe:
With this commit, the time information is still extracted for each file,
even if check_all_file is False. However, other parameters are not checked.
This makes the initialization of OpenPMDTimeSeries ~3x faster than with
check_all_files=True (if particles are present)

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
@@ -14,7 +14,7 @@ import numpy as np
 from .utilities import is_scalar_record, get_shape, get_bpath
 
 
-def read_openPMD_params(filename):
+def read_openPMD_params(filename, extract_parameters=True):
     """
     Extract the time and some openPMD parameters from a file
 
@@ -23,14 +23,17 @@ def read_openPMD_params(filename):
     filename: string
         The path to the file from which parameters should be extracted
 
+    extract_parameters: bool, optional
+        Whether to extract all parameters or only the time
+        (Function execution is faster when extract_parameters is False)
+
     Returns
     -------
     A tuple with:
     - A float corresponding to the time of this iteration in SI units
     - A dictionary containing several parameters, such as the geometry, etc.
+      When extract_parameters is False, the second argument returned is None.
     """
-    params = {}
-
     # Open the file, and do a version check
     f = h5py.File(filename, 'r')
     version = f.attrs['openPMD'].decode()
@@ -38,6 +41,17 @@ def read_openPMD_params(filename):
         raise ValueError(
             "File %s is not supported: Invalid openPMD version: "
             "%s)" % (filename, version))
+
+    # Find the base path object, and extract the time
+    bpath = f[get_bpath(f)]
+    t = bpath.attrs["time"] * bpath.attrs["timeUnitSI"]
+
+    # If the user did not request more parameters, exit now.
+    if not extract_parameters:
+        return(t, None)
+
+    # Otherwise, extract the rest of the parameters
+    params = {}
 
     # Find out supported openPMD extensions claimed by this file
     # note: a file might implement multiple extensions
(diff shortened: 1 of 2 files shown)


soerenjalas:
Nice, Thanks!

ax3l:
great solution, that works.

just a comment: Did you check the performance difference? I suspect that `extract_parameters=True` is actually no real overhead since the main latency is coming in from opening and closing the individual files.

soerenjalas:
For me this is faster by orders of magnitude.

ax3l:
cool, thanks!

soerenjalas:
I tested this for a timeseries with 700 items (320 MB each)

```python
In [2]: %time OpenPMDTimeSeries('.',check_all_files=False)
CPU times: user 527 ms, sys: 55 ms, total: 582 ms
Wall time: 580 ms
Out[2]: <opmd_viewer.openpmd_timeseries.main.OpenPMDTimeSeries at 0x7f03fdacb9d0>
```
The second test is done with a new session, because h5py seems to buffer some data, which makes a second call significantly faster.

```python
In [2]: %time OpenPMDTimeSeries('.',check_all_files=True)
CPU times: user 3.24 s, sys: 696 ms, total: 3.93 s
Wall time: 2min 21s
Out[2]: <opmd_viewer.openpmd_timeseries.main.OpenPMDTimeSeries at 0x7f2d71891a10>
```

ax3l:
huh, now I am [*really curious*](https://docs.python.org/3/library/trace.html) if we can find the slow part ;)

RemiLehe:
From the Python line profiler, it seems that commands that take time are typically distributed all across the `read_openPMD_params` function.

For instance, the command
```python
%lprun -f read_openPMD_params ts = OpenPMDTimeSeries('./example-3d/hdf5/')
```
returns the results [here](https://github.com/openPMD/openPMD-viewer/files/248913/profiling.txt) (look for the column %Time)

ax3l:
cool tool, tried it myself and get to similar results. interesting, if that becomes more urgent we might need to ask the `h5py` folks if it is possible to read all meta data attribute into RAM on file open to speed things up.

Issue #95, opened 2016-04-28
User-friendly installation of the package


RemiLehe:
There are several tasks related to the installation that we should go through:
 - Prepare and test installation/uninstallation with `pip`.
 - Prepare and test installation/uninstallation with `conda`.

In particular, we need to add install instructions in the `README` for both `pip` and `conda`. In the case of `pip`, it would be good to distinguish between:
- installation on a cluster  (which requires `module load` instructions for hdf5 parallel)
- installation on a personal computer (which requires that h5py be properly installed, for instance through pip if possible)

ax3l:
small note: for "personal computer" I would recommend `conda` since that can install a serial hdf5 version required for h5py.

on a cluster `pip` or `conda` works equally but MPI, `parallel hdf5` and `parallel h5py` need to be installed [manually](https://gist.github.com/ax3l/928933f0f2cc80eabdd93d64d98827a1#file-warp-hypnos-profile-L83-L86).

RemiLehe:
Actually, now that I think about it, even if `parallel hdf5` is installed on a given plateform, `openPMD-viewer` will actually not use its parallel features, since nowhere in the code did we include MPI-specific code. I am correct?

ax3l:
For *now* ;)

Yes, as long as we have not parallel backend we can simplify things.

soerenjalas:
I agree, but still I think it's best habit to use the provided mpi and hdf5 versions as the codes will need them anyways

ax3l:
yes that is what I am thinking, too. it's also likely that I want to combine them in the same module + virtualenv combination.

PR #101, opened 2016-05-04
Prepare the code for a PyPI release

RemiLehe:
This pull request makes the package ready for a PyPI release. In particular, I uploaded a version on the [test PyPI](https://testpypi.python.org/pypi) repository, which can be seen [here](https://testpypi.python.org/pypi/openPMD-viewer/0.2.0.2). (Don't pay attention the version number: I had to artificially increase it so that test-PyPI lets me me upload the release, but in the future official release on the real PyPI, this is going to be only version 0.2.0.)

A few comments: 
- Annoyingly, I had to convert the README from Markdown to RST in setup.py because PyPI does not support Markdown.
- I also had to change some classifiers in the setup.py because PyPI only supports certain pre-defined classifiers.
- I added documentation on how to create a release in CONTRIBUTING.md

NB: In order to close issue #95, I still need to create a conda release, and to update the install instructions in README.md


ax3l:
how great, will review! :)

setup.py
```diff
@@ -1,41 +1,46 @@
...
-      platforms='any',
       classifiers=[
           'Programming Language :: Python',
           'Development Status :: 4 - Beta',
           'Natural Language :: English',
           'Environment :: Console',
-          'Environment :: Jupyter',
           'Intended Audience :: Science/Research',
-          'License :: BSD-3-Clause-LBNL',
+          'License :: OSI Approved :: BSD License',
```
ax3l: yes I know you like to change this line (and `license` above) to a pre-defined identifier but it's actually not 100% correct.
the LBNL (and LLNL) BSD-style licenses are a 3-Clause BSD license *but* with an extensions reserving them to use improvements for a certain amount of time. I think we should better be transparent about this before someone complains :)

Also, it will still work ok enough. See Forthon:
- [pypi](https://pypi.python.org/pypi/Forthon)
- [badge](https://github.com/dpgrote/Forthon/blob/master/README.md)

RemiLehe:
Ok, I updated the information to make it correct. I also removed the license information from the classifiers (as it is done for instance also for Forthon), since PyPI parses the classifiers and only allows certain strings (and 'License :: BSD-3-Clause-LBNL' is not one of them: see [here](https://pypi.python.org/pypi?%3Aaction=list_classifiers) for a list of allowed classifiers)

setup.py
```diff
@@ -1,41 +1,45 @@
...
       setup_requires=['pytest-runner'],
-      platforms='any',
       classifiers=[
           'Programming Language :: Python',
           'Development Status :: 4 - Beta',
           'Natural Language :: English',
           'Environment :: Console',
-          'Environment :: Jupyter',
           'Intended Audience :: Science/Research',
-          'License :: BSD-3-Clause-LBNL',
```
ax3l: @RemiLehe that is not a problem, you can leave it in. We also use it in [Forthon](https://github.com/dpgrote/Forthon/blob/901b43f056e19dabf7c2ceeabaee0d2bebf6bb1d/setup.py#L105) and the provided list in pypy is just a "predefined"/"known" list but does not disallow to take other keywords.

setup.py
```diff
@@ -1,41 +1,45 @@
...
       install_requires=install_requires,
       tests_require=['pytest', 'jupyter'],
       setup_requires=['pytest-runner'],
-      platforms='any',
       classifiers=[
           'Programming Language :: Python',
           'Development Status :: 4 - Beta',
           'Natural Language :: English',
           'Environment :: Console',
-          'Environment :: Jupyter',
```
ax3l: you can add own categories in pypy, I think that might be a good one

setup.py
```diff
@@ -1,41 +1,45 @@
...
-      url='git@bitbucket.org:berkeleylab/opmd_viewer.git',
+      license='BSD-3-Clause-LBNL',
+      url='https://github.com/openPMD/openPMD-viewer.git',
       packages=find_packages('./'),
       package_data={'opmd_viewer': ['notebook_starter/*.ipynb']},
       scripts=['opmd_viewer/notebook_starter/openPMD_notebook'],
       install_requires=install_requires,
       tests_require=['pytest', 'jupyter'],
       setup_requires=['pytest-runner'],
-      platforms='any',
```
ax3l: why did you remove that? the openPMD-viewer is platform independent code :)

setup.py
```diff
@@ -1,41 +1,45 @@
...
           'Development Status :: 4 - Beta',
           'Natural Language :: English',
           'Environment :: Console',
-          'Environment :: Jupyter',
           'Intended Audience :: Science/Research',
-          'License :: BSD-3-Clause-LBNL',
           'Operating System :: OS Independent',
           'Topic :: Scientific/Engineering :: Physics',
           'Topic :: Scientific/Engineering :: Visualization',
-          'Topic :: Database :: Front-Ends',
```
ax3l: I found this matching because it's a front-end for a file-based database (openPMD) with a well defined structure

ax3l:
@RemiLehe shall quick-check on Monday about the last inline points?

RemiLehe:
Sure, let's do this!

setup.py
RemiLehe: OK, I'll keep it.

setup.py
RemiLehe: Ok, I'll keep it.

RemiLehe:
Here is for instance the error that I get when trying to upload the code with `'License :: BSD-3-Clause-LBNL'`

```
twine upload dist/* -r pypitest
Uploading distributions to https://testpypi.python.org/pypi
Enter your password: 
Uploading openPMD_viewer-0.2.0.5-py2.py3-none-any.whl
HTTPError: 400 Client Error: Invalid classifier "License :: BSD-3-Clause-LBNL" for url: https://testpypi.python.org/pypi
```

RemiLehe:
As a result, I'll do all the requested modifications, except those that are refused by PyPI

ax3l:
ok, sure! Thanks for testing that!

PR #104, opened 2016-05-19
Import numpy and matplotlib in the template notebook

RemiLehe:
I updated the template notebook that one gets when typing `openPMD_notebook`.
In particular, I import numpy and matplotlib by default, since I always find that I need to import them in practical situations.


soerenjalas:
Nice idea! Also always the first thing I do when creating a notebook.

Issue #103, opened 2016-05-12
positionOffset: int32 and untiSI


ax3l:
When trying to read a PIConGPU file with the current `dev` branch as of 135809c145c0e70f3d495ce1d612df8633897b7e reading the position via
```python
y,x = ts.get_particle( var_list=['y','x'], t=0, species='e', plot=True, nbins=300, vmax=5e7)
```
leads to:
```
openPMD-viewer/opmd_viewer/openpmd_timeseries/main.pyc in get_particle(self, var_list, species, t, iteration, select, output, plot, nbins, **kw)
    239         for quantity in var_list:
    240             data_list.append(read_species_data(
--> 241                 file_handle, species, quantity, self.extensions))
    242         # Apply selection if needed
    243         if select is not None:

openPMD-viewer/opmd_viewer/openpmd_timeseries/data_reader/particle_reader.pyc in read_species_data(file_handle, species, record_comp, extensions)
     72     # - Return positions in microns, with an offset
     73     if record_comp in ['x', 'y', 'z']:
---> 74         offset = get_data(species_grp['positionOffset/%s' % record_comp])
     75         data += offset
     76         data *= 1.e6

/openPMD-viewer/opmd_viewer/openpmd_timeseries/data_reader/utilities.pyc in get_data(dset, i_slice, pos_slice)
     92     # Scale by the conversion factor
     93     if dset.attrs['unitSI'] != 1.0:
---> 94         data *= dset.attrs['unitSI']
     95 
     96     return(data)

TypeError: Cannot cast ufunc multiply output from dtype('float64') to dtype('int32') with casting rule 'same_kind'
```

That is likely do to the recent changes in #81 #85 and caused by our internal storage of `positionOffset` as an `int32` :)

The result of `positionOffset[()]` (int32) and `unitSI` (float64) should of cause be promoted to `float64` if `unitSI` != `1.0`. Since `*=` is an [in place operation](http://docs.scipy.org/doc/numpy-dev/user/quickstart.html#basic-operations) we can just add a reallocation & cast beforehand if necessary:
```python
if dset.attrs['unitSI'] != 1.0:
    if data.dtype != np.float64:
        data = data.astype(np.float64)
    data *= dset.attrs['unitSI']
```

[...]

Issue #99, opened 2016-05-03
axisLabels not Evaluated


ax3l:
Plotting e.g., 2D fields right now does not add the correct labels to the axis (e.g., they are `x`, `z` as in warp).

For that to change, we just need to read the [`axisLabels` attribute](https://github.com/openPMD/openPMD-standard/blob/1.0.0/STANDARD.md#required-attributes-for-each-mesh-record).

RemiLehe:
Should we close this issue?

RemiLehe:
This was fixed by PR #102

PR #108, opened 2016-06-16
Install instructions

RemiLehe:
I changed the installation instructions to make them compatible with the future `pip` distribution.

I also put the **Usage** section first in the README, in order to encourage the users to go look the tutorials. (Before this, they had to read through the installation instructions for various systems, before learning about the existence of the tutorials. I guess most users would stop reading before then.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,7 @@ before_install:
 
 # Install packages
 install:
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION h5py
-  - pip install -r requirements.txt
+  - conda install --yes python=$TRAVIS_PYTHON_VERSION --file requirements.txt
   - python setup.py install
   - pip install wget
   - pip install pyflakes pep8 jupyter
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,8 @@ before_install:
 
 # Install packages
 install:
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION --file requirements.txt
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION jupyter
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION pyflakes
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION pep8
+  - conda install --yes python=$TRAVIS_PYTHON_VERSION scipy matplotlib h5py jupyter
+  - conda install --yes python=$TRAVIS_PYTHON_VERSION pyflakes pep8
   - python setup.py install
 
 before_script :
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ before_install:
 install:
   - conda install --yes python=$TRAVIS_PYTHON_VERSION scipy matplotlib h5py jupyter
   - conda install --yes python=$TRAVIS_PYTHON_VERSION pyflakes pep8
+  - pip install wget
   - python setup.py install
 
 before_script :
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,11 @@ before_install:
 
 # Install packages
 install:
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION scipy matplotlib h5py jupyter
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION pyflakes pep8
-  - pip install wget
+  - conda install --yes python=$TRAVIS_PYTHON_VERSION h5py
+  - pip install -r requirements.txt
   - python setup.py install
+  - pip install wget
+  - pip install pyflakes pep8 jupyter
 
 before_script :
   # static code analysis
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,10 @@ git clone git@github.com:<YourUserLogin>/openPMD-viewer.git
 ```
 git checkout dev
 ```
+and install it
+```
+python setup.py install
+```
 
 - Start a new branch from the development branch, in order to
 implement a new feature. (Choose a branch name that is representative of the
@@ -52,6 +56,7 @@ ensure that your code complies with some standard style conventions.
   ```
   - Make sure that the tests pass
   ```
+  python setup.py install
   python setup.py test
   ```
   (Be patient: the `test_tutorials.py` can take approx. 20 seconds if
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,7 +54,7 @@ ensure that your code complies with some standard style conventions.
   pyflakes opmd_viewer
   pep8 --ignore=E201,E202,E122,E127,E128,E131 opmd_viewer
   ```
-  - Make sure that the tests pass
+  - Make sure that the tests pass (please install `wget` and `jupyter` before running the tests: `pip install wget jupyter`)
   ```
   python setup.py install
   python setup.py test
--- a/README.md
+++ b/README.md
@@ -30,7 +30,9 @@ the API and the interactive GUI. You can view these notebooks online
 
 You can also download and run these notebooks on your local computer
 (when viewing the notebooks with the above link, click on `Raw` to be able to
-save them to your local computer).
+save them to your local computer). In order to run the notebook on
+your local computer, please install `openPMD-viewer` first (see
+below), as well as `wget` (`pip install wget`).
 
 ### Notebook quick-starter
 
@@ -41,7 +43,7 @@ browser**. To use this executable, simply type in a regular terminal:
 
 `openPMD_notebook`
 
-(This executable is installed by default.)
+(This executable is installed by default, when installing `openPMD-viewer`.)
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -12,77 +12,71 @@ This package contains a set of tools to load and visualize the
 contents of a set of [openPMD](http://www.openpmd.org/#/start) files
 (typically, a timeseries).
 
-## Installation
-
-### Basic installation
-
-To install this package :
-
-- Clone this repository using `git`
-```
-git clone https://github.com/openPMD/openPMD-viewer.git
-```
-
-- `cd` into the directory `openPMD-viewer` and run
-```
-python setup.py install
-```
-
-### Installing the interactive GUI
-
-The **interactive GUI** for IPython Notebook is not
-operational by default.
-
-This is because it requires dependencies that may be difficult to
-install on some systems. If you wish to have the interactive GUI
-working, install the
-[IPython Notebook](http://ipython.org/notebook.html)
-(now part of the [Jupyter project](http://jupyter.org/)) by hand:
-
-`conda install jupyter` (for the
-[Anaconda](https://www.continuum.io/downloads)
-distribution) or `pip install jupyter` (for the other Python distributions)
-
-NB: For [NERSC](http://www.nersc.gov/) users, it is not necessary to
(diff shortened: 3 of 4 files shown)


.travis.yml
```diff
@@ -30,10 +30,8 @@ before_install:
 
 # Install packages
 install:
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION --file requirements.txt
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION jupyter
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION pyflakes
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION pep8
+  - conda install --yes python=$TRAVIS_PYTHON_VERSION scipy matplotlib h5py jupyter
```
ax3l: this change is not ideal: we want to make sure `requirements.txt` is up-to-date, so let us install with that file if possible

README.md
```diff
@@ -12,77 +12,71 @@ This package contains a set of tools to load and visualize the
...
+pip install openPMD-viewer
+```
+The first line installs the dependencies, while the second line
+installs `openPMD-viewer` itself
+
+### Installation on a remote scientific cluster
+
+If you wish to install the `openPMD-viewer` on a remote scientific
+cluster, please make sure that packages `numpy`, `scipy` and `h5py`
+are available in your environment. This is typically done by a set of
```
ax3l: let us add the note that those are up-to-date in `requirements.txt`

.travis.yml
ax3l: aktually the way it was written was good:
- install basic dependencies via `requirements.txt`
- install optional dependency `jupyter` for gui tests in travis
- install checkers `pyflakes` and `pep8`

if you don't had something else in mind, pls revert the changes in this file

README.md
```diff
@@ -12,77 +12,71 @@ This package contains a set of tools to load and visualize the
...
+
+### Installation on a local computer, using Anaconda
+
+On a local computer, it is recommended to install the [Anaconda
+distribution](https://www.continuum.io/downloads), which enables an easy
+installation of the dependencies of `openPMD-viewer`.
+
+Once Anaconda is installed, type
+```
+conda install scipy matplolib h5py jupyter
```
ax3l: motivate to use `requirements.txt`

ax3l:
added some inline comments. `requirements.txt` is the typical place to put all required dependencies to not run into documentation-updating-madness as soon as we add or remove further dependencies.

README.md
RemiLehe: The problem is that users will not have access to `requirements.txt` when using pip. (It is not a `git clone`, they do not see the source).

Now of course, pip **will still have knowledge of the requirements from `requirements.txt`** (because we open and parse `requirements.txt` in setup.py), and so if the user skips `conda install scipy matplolib h5py jupyter` before using `pip`, `pip` will still try to install those packages (again, because they are in the requirements), but with `pip` and not `conda`. The whole problem is that the installation of `h5py` with `pip` **does not work** while the `conda` installation works. 

Therefore the user should at least type `conda install h5py`. Since the user types this anyway, I think it is good that he also installs `scipy`, `matplotlib` and `jupyter` with `conda` so that everything is consistent. (and `conda` tends to be a much better package manager than `pip`).

I know this scheme is not ideal, and I would not have used if it wasn't for that installation problem with `pip` and `h5py`. Do you have any suggestion to make it better (again, keep in mind that the user will not have access to requirements.txt when downloading the code with `pip`)

.travis.yml
RemiLehe: Yeah, the aim of the above line was to create a test of the installation which is **as close as possible** to what the user types: see the other comment below. (except that we do not use `pip install openPMD-viewer` but `python setup.py install` to install the latest version of the code.)

Also, even with the above commands, `requirements.txt` will still be used (since setup.py is reading it when doing `python setup.py install`). The only reason why I was adding `conda install --file requirements.txt` is so that the requirements get installed with `conda` instead of `pip`. (again, see remark below)

ax3l: > Yeah, the aim of the above line was to create a test of the installation which is as close as possible to what the user types [...]

I know, but the user *should* type:
```
conda install --yes --file requirements.txt
```

ax3l: when installing from scatch.

if requirements.txt is read and installed when typing `python setup.py install` then we should not write this at all.

ax3l: > The only reason why I was adding conda install --file requirements.txt is so that the requirements get installed with conda instead of pip. (again, see remark below).

why should we mix the two install methods in travis?

README.md
ax3l: > The problem is that users will not have access to requirements.txt when using pip

When using pip, the package install will determine the dependency of `pip install openPMD-viewer` via the packages `requirements.txt` and install them automatically for them.

.travis.yml
ax3l: reading below... will follow up in a sec

README.md
ax3l: > The whole problem is that the installation of h5py with pip does not work while the conda installation works.

the h5py pip install is tricky for parallel installs indeed but works for me in normal setups.

you could alternatively just install the pip package `openPMD-viewer` via `conda` on OSX so the dependencies will be auto-installed with and by conda, too.

ax3l: I would generally advise the user to go for one install method, pip or conda based, when starting.

so just go for
```bash
conda install openPMD-viewer
```

below in that case and it will find and install the pip described dependencies but *through* conda.

.travis.yml
ax3l: travis-world: in travis we do not install the package but we test source-branches. in that case, we can just stay with the description as it was.

if we want to *additionally* test the on-demand build packages with *each commit* then we should extent the build matrix to that use case. I did so in [Forthon](https://github.com/dpgrote/Forthon/blob/8eb122f031adc17c4f6f925d249badcc2a718ccf/.travis.yml#L26-L29) and I would like to do that in an independent PR because it will be a bit more, independent code.

README.md
ax3l: to be honest: the module load python packages do not work 95% of the time due to ABI incompatebilities with various compilers that are common.

Due to that many sysadmins deprecate to even provide more than a plain python and encourage to go for venvs in user mode.

RemiLehe:
@ax3l : Ok thanks a lot for the feedback. I initially wanted to delay the preparation of a conda install for later, but I now agree with you that it is needed for consistency here. I'll update the pull request within the next few days. 

I might drop the 'wget' requirement as it is not in the standard conda channels, and is not needed unless you run the tutorials locally (which users are much less likely to do when installing via pip/conda as they will, again, not have access to the source)

ax3l:
Sounds great, thank you! :)

Just for the record, installing `h5py` with pip will soon be easier: https://github.com/h5py/h5py/pull/721

ax3l:
> I might drop the 'wget' requirement as it is not in the standard conda channels

but we won't be in the standard channels either, or will we?

README.md
RemiLehe: By the way:

> the h5py pip install is tricky for parallel installs indeed (h5py/h5py#542, hand-ish) but works for me in normal setups.

I think that even with serial h5py, the installation process is often likely to crash on a fresh system, when trying to install it with `pip`.

Try for instance:
```
docker run -it ubuntu
apt-get update
apt-get install python-pip
pip install h5py
```

ax3l: oh sad! what is the problem in such a case?

ax3l: uh that is light weight, downloaded!

ax3l: So tried it (with `-v` to follow the progress of pip).
`h5py` first pulls `numpy` and `Cython` automatically, each compiling forever but worked. Then it complains about a missing hdf5 library, because pip only maintains python modules (requirement is a `module load hdf5` or `apt-get install`).

The install dir on ubuntu is [super annoying](http://stackoverflow.com/questions/30244669/installing-h5py-on-ubuntu-15-04) causing many projects that are not cmake based to struggle since debian/ubuntu like to split and re-distribute files all over the place, such as `/usr/include/hdf5/serial/` in that case for the serial hdf5 headers (without symlinks in /usr/include ...).

My serial hdf5, ubuntu docker install was then, using the wrapper compilers:
```bash
docker run -it ubuntu
apt-get update
apt-get upgrade -y
apt-get install -y python-pip libhdf5-dev

HDF5_DIR=/usr/lib/x86_64-linux-gnu/hdf5/serial/ pip install h5py
```

RemiLehe: Ok thanks for letting me know how to fix it. 

Regarding the overall idea of this pull request. I had a look at the [conda build](http://conda.pydata.org/docs/build_tutorials.html) instructions, and it seems that setting up a conda build would be **much easier** (esp. to maintain) if we already have a working pip release. 

Therefore, I was thinking to do a first release (and upload it to PyPI), with only pip instructions. And later do a second release with a conda build and conda instructions. Is it okay with you?

If yes, should I include the above Debian-specific instructions that you figured out.

RemiLehe:
Yes, openPMD-viewer will not be in the standard channels, but still, in order to build it, it is easier to use only code that is in standard channels. There are [workarounds](http://conda.pydata.org/docs/build_tutorials/pkgs.html#troubleshooting), but I am not sure that it is worth it in the case of wget. (Again, it is only needed if the user wants to run the tutorials locally, and in this case we can also tell the user to do `pip install wget`)

RemiLehe:
If you agree with the last remark, I'll update the PR to:
- Remove mentions of conda (for now)
- Remove wget

Is it okay?

ax3l:
Sounds great, let's do it that way!

RemiLehe:
Ready for merging!

PS: Please also have a look at issues #99 and #103 : once they are fixed, we can do the PyPI release.

.travis.yml
```diff
@@ -31,10 +31,9 @@ before_install:
 # Install packages
 install:
   - conda install --yes python=$TRAVIS_PYTHON_VERSION --file requirements.txt
```
ax3l: if wget is now in `pip` we should be able to install the `requirements.txt` via `pip install -r requirements.txt` too.

CONTRIBUTING.md
```diff
@@ -52,6 +56,7 @@ ensure that your code complies with some standard style conventions.
   ```
   - Make sure that the tests pass
   ```
+  python setup.py install
   python setup.py test
```
ax3l: you should mention that running `test` requires the additional install of `pip install wget`

ax3l: maybe there is also a designated place in `setup.py` for test-only requirements?

ax3l: oh [there is](http://stackoverflow.com/questions/4734292/specifying-where-to-install-tests-require-dependencies-of-a-distribute-setupto#7747140)! :) just add `wget` to our `tests_require` :)

ax3l: there is also a *good practices* blog post on `pytest`: https://pytest.org/latest/goodpractices.html

RemiLehe: I is very strange, but even though I added `wget` to the `tests_requires` and pip does **download** it, Python is not able to find it when running the tests: see https://travis-ci.org/RemiLehe/openPMD-viewer/jobs/139221627

For now, I suggest just telling developers to install `wget` inside `CONTRIBUTING.md` (`pip install wget` works well). Let's try to have a PyPI version rather soon, and then we can improve it later.

ax3l:
thank you for the many updates! :sparkles:

RemiLehe:
Thanks for being a careful and dedicated reviewer :)

PR #111, opened 2016-06-27
Bad evaluation of the waist

RemiLehe:
# Summary

I found that the method `get_laser_waist` gives quite a bad estimation of the waist for 2 reasons:
- The extraction of a transverse slice (from a 2D map of the envelope) uses `np.amax`
- The waist is evaluated with an RMS, instead of doing a Gaussian fit

I corrected those two limitations. Below is a more in-depth explanation of the problems with the evaluation of the waist.

# Explanation

## The problem

I recently did a simulation which involved self-focusing. Here is how the evolution of the waist and a0 look like, when evaluated with openPMD-viewer:
waist_before
However, something is quite shocking: the a0 changes a lot compared to the vacuum prediction (indicating strong self-focusing) while the waist changes very little compared to the vacuum prediction (indicating a low amount of self-focusing).

## Extraction of a transverse slice

Part of the problem is that the transverse slice is extracted by using `np.amax( ... , axis=1)`. This searches for the position of the maximum of the fields **independently for each radial position**. Therefore, the extracted profile is taken along a **curved path** that follows the maximum of the field (blue curve in the top panel below), instead of being taken along a straight transverse slide (green curve in the top panel below). Thus -- especially for peanut-shaped laser pulse -- the extracted field profile can look quite different from the  field profile in a straight transverse slice (see bottom panel), and of course the evaluation of the waist will be affected.
figure_1
I corrected this issue by taken a normal, straight transverse slice. 

Here is the new evolution of the waist with this method. It is better, but not quite credible yet. 
waist_better

## Fitting of a Gaussian

The remaining problem is that the field profile (green curve above) has wings, and thus the current RMS calculation of the waist will give a strong weight to these wings and overestimate the waist. By contrast, fitting a Gaussian pulse gives a much better estimation of the waist (see figure below with a Gaussian fit in dashed line)

fit

Here is the corresponding evolution of the waist, when evaluated with the fitting method. I looks much more credible.

with_fit

In the code, I kept the option to use the RMS method ,but the user needs to specify it. By default, the fitting method is used.

### DIFF
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -694,13 +694,14 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
                                                 pol=pol, index='all',
                                                 slicing_dir=slicing_dir,
                                                 theta=theta)
-
-        # Find the maximum of the envelope along the transverse axis
-        trans_max = np.amax(field, axis=1)
+        # Find the indices of the maximum field, and
+        # pick the corresponding transverse slice
+        _, iz_max = np.unravel_index( np.argmax( field ), dims=field.shape )
+        trans_slice = field[ :, iz_max ]
         # Get transverse positons
         trans_pos = getattr(info, info.axes[0])
         # Calculate standard deviation
-        sigma_r = wstd(trans_pos, trans_max)
+        sigma_r = wstd(trans_pos, trans_slice)
         # Return the laser waist = sqrt(2) * sigma_r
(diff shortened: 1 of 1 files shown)


MKirchen:
Thanks for these detailed explanations. I think Sören will merge this asap.

soerenjalas:
Thanks for the thorough analysis of the problem. I agree that I didn't think this through completely. I hope it didn't cost you too many hours of work. Thanks for the fix!

RemiLehe:
No problemo. Thanks for merging!

PR #112, opened 2016-06-29
Close #110 : Add the resolution (e.g. dx, dz) to the FieldMetaInformation object

### LINKED ISSUE #110: Add the resolution (e.g. dx, dz) to the FieldMetaInformation object
It would be good if the user can extract the resolution easily.

soerenjalas:
As discussed in #110, the grid resolution can be queried with this PR via `FieldMetaInformation.dz`.


RemiLehe:
Great! Thanks a lot!

PR #113, opened 2016-06-29
Close #103: positionOffset: int32 and untiSI

### LINKED ISSUE #103: positionOffset: int32 and untiSI

RemiLehe:
Close #103 This attempts to fix the above mentioned issues with PIConGPU files. 
@ax3l : Could you check that it does fix the issue by testing it one of the files that was raising the error you reported?

Thanks!

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/utilities.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/utilities.py
@@ -89,7 +89,7 @@ def get_data(dset, i_slice=None, pos_slice=None):
         elif pos_slice == 2:
             data = dset[:, :, i_slice]
 
-    # Scale by the conversion factor
+    # Convert to float and scale by the conversion factor
     if data.dtype != np.float64:
         data = data.astype(np.float64)
     if dset.attrs['unitSI'] != 1.0:
--- a/opmd_viewer/openpmd_timeseries/data_reader/utilities.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/utilities.py
@@ -90,6 +90,8 @@ def get_data(dset, i_slice=None, pos_slice=None):
             data = dset[:, :, i_slice]
 
     # Scale by the conversion factor
+    if data.dtype != np.float64:
+        data = data.astype(np.float64)
     if dset.attrs['unitSI'] != 1.0:
         data *= dset.attrs['unitSI']
 


RemiLehe:
Ready for merging! (The previous error in the automated tests was simply due to a transient OSX failure on TravisCI)

ax3l:
sorry for the delay! I tested this branch and it works like a charm! thank you! :sparkles:

Issue #115, opened 2016-07-01
Add fit to get_main_frequency 


soerenjalas:
Currently the method `get_main_frequency` in `LpaDiagnostics` simply searches for the maximum in the FFT. This produces quite unusable data when studying e.g. the redshift of the pulse in plasma. As you can see below the value changes quite unsteadily.
![freq](https://cloud.githubusercontent.com/assets/10922951/16520241/28fd918c-3f8f-11e6-8219-75fdb9840e94.png)

I think by fitting the spectrum one could probably get better results as more data is taken into account.
The question is if a Gaussian fit is sufficient for this purpose as the pulse can change quite violently or people might use exotic pulses. I can work out a PR so we could test if this method gives better results. @RemiLehe any thoughts on this?

RemiLehe:
OK, I see your point. I think fitting a Gaussian is fine, since for a Gaussian laser the spectrum is indeed a Gaussian (at least before significant pulse evolution). As for the other methods, it would be good to give the user the possibility to still use the maximum of the spectrum if they wish.

soerenjalas:
Closed in #116

PR #116, opened 2016-07-05
Add option to calculate central laser frequency by Gaussian fitting

soerenjalas:
As proposed in #115, I added the option to use a Gaussian fit ( love scipy.optimize! 😊) in the `get_main_frequency` method. The user has now the option to use the old method where only the frequency with the highest intensity in the spectrum is returned or use the new method by giving the option `method='max'` or `method='fit'`.

I compared the two cases with some LWFA simulation data I had (pulse from experiment, so spectrum not necessarily Gaussian):
#### With `method='max'`

It seems that the new method creates a little more compatible data, when looking at the pulse evolution. However, I noticed that the fit seems to be quite sensitive to the start value of the width of the Gaussian.
When I used the standard deviation of the spectrum the fit always resulted in really slim peaks around the maximum of the spectrum and gave the same result as the `method='max'` version. Now I use the width of the full spectrum as a start value and it seems to give good results.

All in all I think we should test this on a few more cases before adding it to a release version just to see if it behaves properly in all cases.

### DIFF
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -515,7 +515,7 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
             # Guess start values for fit
             f0 = omega0
             fmax = np.amax( spectrum )
-            fsigma = wstd(spectrum, info.omega)
+            fsigma = info.omegamax - info.omegamin
             params, _ = curve_fit( gaussian_profile, info.omega,
                                    spectrum, p0=[f0, fmax, fsigma])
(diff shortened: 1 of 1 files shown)


RemiLehe:
Agreed. I'll try to use it whenever I have the occasion. 
Also, would you like me merge this with `dev` or should we wait for it to be tested on a number of different cases?

soerenjalas:
Actually I don't really expect problems with the fit, I just wanted to have this said in case there are any problems. I think you can merge if you don't have any objections.

RemiLehe:
Ok, thanks. I'm okay with the code :) This will be very useful!

PR #117, opened 2016-07-07
Added `plot` option to LpaDiagnostics methods

soerenjalas:
To make the user experience a little more consistent I added `plot=True/False` options to all the methods in `LpaDiagnostics` where it makes sense. Also to make it consistent with the main API I added a title with the current iteration and time to the `get_spectrum` plot.

### DIFF
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -92,7 +92,7 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
         return( mean_gamma, std_gamma )
 
     def get_sigma_gamma_slice(self, dz, t=None, iteration=None, species=None,
-                              select=None):
+                              select=None, plot=False, **kw):
         """
         Calculate the standard deviation of gamma for particles in z-slices of
         width dz
@@ -119,6 +119,12 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
             'x' : [-4., 10.]   (Particles having x between -4 and 10 microns)
             'z' : [0, 100] (Particles having x between 0 and 100 microns)
 
+        plot : bool, optional
+           Whether to plot the requested quantity
+
+        **kw : dict, otional
+           Additional options to be passed to matplotlib's `plot` method
+
         Returns
         -------
         A tuple of arrays:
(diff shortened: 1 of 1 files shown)


RemiLehe:
Sorry for taking so long before finally having some time to review this.
I like the pull request very much: this will be very useful. 
I have a few corrections (see comments), but once they are implemented, the PR is good to be merged :)

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -422,6 +446,24 @@ def get_laser_envelope( self, t=None, iteration=None, pol=None, m='all',
         if index != 'all':
             info.restrict_to_1Daxis( info.axes[1] )
 
+        # Plot the result if needed
+        if plot:
+            iteration = self.iterations[ self.current_i ]
+            time_fs = 1.e15 * self.t[ self.current_i ]
+            if index != 'all':
+                plt.plot( info.z, envelope, **kw)
+                plt.ylabel('$E_%s \;(V/m)$' % pol,
```
RemiLehe: Actually, info.z is in SI (meters in this case). So, I would use `1.e6*info.z` here, since the labels indicate microns.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -422,6 +446,24 @@ def get_laser_envelope( self, t=None, iteration=None, pol=None, m='all',
...
+                           fontsize=self.plotter.fontsize)
+            else:
+                plt.imshow( envelope, extent=info.imshow_extent,
```
RemiLehe: Similarly, `info.imshow_extent` is also in SI. Please use `1.e6*info.imshow_extent`.

Btw, Sorry for these inconsistencies in openPMD-viewer: some quantities are in microns and some other are in SI. My aim is to uniformize all of this in the future.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -92,7 +92,7 @@ def get_mean_gamma( self, t=None, iteration=None, species=None,
         return( mean_gamma, std_gamma )
 
     def get_sigma_gamma_slice(self, dz, t=None, iteration=None, species=None,
-                              select=None):
+                              select=None, plot=False, **kw):
         """
         Calculate the standard deviation of gamma for particles in z-slices of
         width dz
```
RemiLehe: This is somewhat unrelated, but could you add, in the docstring, that `dz` should be passed in microns? When testing, I mistakenly tried to pass it in meters (i.e. `dz=1.e-6`) and the function just hung (too many bins).

soerenjalas:
Sorry for the delay, but I implemented your suggestions. If you don't see anything else you can go ahead and merge.

RemiLehe:
Thanks for implementing the changes !

PR #119, opened 2016-07-07
Initialize avail_circ_modes to empty list instead of None for Cartesian data

### LINKED ISSUE #118: avail_circ_modes set to None causes OpenPMDTimeSeries.slider() to fail
It appears that for Cartesian datasets, `avail_circ_modes` will be set to `None` in [params_reader.py](https://github.com/openPMD/openPMD-viewer/blob/dev/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py#L99), but this appears to cause an error when calling `OpenPMDTimeSeries.slider()` for interactive exploration in an IPython (`4.2.1-aed0eae`) notebook with widgets.  This is replicated for the [example datasets](https://github.com/openPMD/openPMD-example-datasets)

Replacing `avail_circ_modes` with `[]` as needed after creating the `OpenPMDTimeSeries` object seems to yield the correct behavior:

```
import opmd_viewer

ts = opmd_viewer.OpenPMDTimeSeries('./example-3d/hdf5/')

%matplotlib inline
if ts.avail_circ_modes is None:
    ts.avail_circ_modes = []
ts.slider()
```

Unless I've misunderstood something about this object, using the empty list `[]` instead of `None` at [params_reader.py:L99](https://github.com/openPMD/openPMD-viewer/blob/dev/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py#L99) will resolve the issue without the need for intervention on the user's part.

jgerity:
Resolves #118, per @RemiLehe's request

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
@@ -96,7 +96,7 @@ def read_openPMD_params(filename, extract_parameters=True):
                 params['geometry'] = "2dcartesian"
             elif dim == 3:
                 params['geometry'] = "3dcartesian"
-            params['avail_circ_modes'] = None
+            params['avail_circ_modes'] = []
         # For each field, check whether it is vector or scalar
         # Store the information in a dictionary
         params['avail_fields'] = {}


RemiLehe:
Great! Thanks a lot for the contribution!!

PR #122, opened 2016-07-22
Revert to evaluating laser frequency with max

RemiLehe:
When testing the new `get_main_frequency` (with method=`fit`), I encountered several problems. As @soerenjalas already warned, the fit is not very robust. In some cases, I found that it returned negative frequencies... This is quite inconvenient, especially as this function is called e.g. in `get_a0`.

I propose that we keep the `fit` method (as it is also quite convenient in some cases, as shown by @soerenjalas in his pull request), but that the default is to use `max`, which is more robust.


soerenjalas:
Okay, thanks for pointing that out.

PR #127, opened 2016-10-20
Perform tests within a conda environment

RemiLehe:
This PR modifies the Travis CI so that they run in a conda environment.

The main motivation is to ensure that the python which is used to run the test is indeed the intended python (i.e. python 2.7 and python 3.5 respectively, for the two set of tests that we run)


ax3l:
looks good to me!

PR #129, opened 2016-11-29
Close #128: Added the __version__ attribute

### LINKED ISSUE #128: Module: Version Attribute Missing?
Hi @RemiLehe,

Does the main package of the viewer lack a `__version__` attribute, or am I missing it?

```python
import opmd_viewer

print(opmd_viewer.__version__)
# AttributeError: 'module' object has no attribute '__version__'
```

We might want to add it, there is are two PEP about it :)
https://github.com/ornladios/ADIOS/issues/61#issue-134347515

RemiLehe:
I added the `__version__` attribute.
Note that this creates a duplication (since the version number is also indicated in the [`setup.py`](https://github.com/openPMD/openPMD-viewer/blob/master/setup.py)).

One usual way to avoid the duplication is to **import** the package itself (and its attribute `__version__`) in the `setup.py`. (This is for instance how we do it in fbpic: https://github.com/fbpic/fbpic/blob/dev/setup.py). However, this would not work in the case of openPMD-viewer, because we import other objects in the `__init__.py`, and this import statement would cause a crash if the dependencies (e.g. numpy, h5py, etc.) are not installed already. So for instance, if we import `opmd_viewer` in the `setup.py`, and then do `python setup.py install`, the command would crash **before** installing the dependencies of `opmd_viewer`.

Does this make sense? Are you fine with the current solution (although it involves a duplication)?


ax3l:
Thanks!

I know that common problem and I think the usual way to solve it is to store the version in a zero-dependency file separately and once and only include & use that file containing only the version in both in `setup.py` and `__init__.py`.

RemiLehe:
OK, that makes sense. I updated the PR. Note that I cannot **import** `__version__` from the `setup.py`, otherwise the `__init__.py` will be executed. Instead I have to use a small hack where I **execute** the file `__version__.py` from the `setup.py`.

ax3l:
awesome, thank you! :sparkles:

PR #139, opened 2017-01-17
Close #137: Add particle tracker in openPMD-viewer

### LINKED ISSUE #137: Add particle tracker in openPMD-viewer
For one of the sets of simulations that I am currently doing, I would need to be able to track particles in postprocessing. 

Assuming that the `id` of the particles have been stored in the openPMD data, then this should be possible with openPMD-viewer, if we add this feature. 

I would definitely volunteer for this, but I would like us to first agree on the API. I thought about introducing a `ParticleTracker` object. This object would be used in the following way, for instance in order to select electrons having a longitudinal momentum between 10 and 100 at iteration 2000, and then fetching the positions of these electrons at iteration 30000 and 50000.
```python
ts = OpenPMDTimeSeries( 'some/directory' )

# Select particles at iteration 2000
pt = ParticleTracker( ts, species='electrons', select={'uz':[10,100]}, iteration=2000 )

# Fetch them at iteration 30000 and 50000
x1, y1, z1 = ts.get_particle( ['x', 'y', 'z'], particle_tracker=pt, iteration=30000 )
x2, y2, z2 = ts.get_particle( ['x', 'y', 'z'], particle_tracker=pt, iteration=50000 )
```
Also, in this case x1, y1, etc. would be arrays that have the same length as the initial number of particles selected at iteration 2000, and the same tracked particle would have the same position within the array x1 and x2 for instance. Also, there would be NaNs in x1 if some of the particles were present at iteration 2000 but not anymore at the iteration 30000.

@ax3l @soerenjalas @MKirchen @jgerity : Does the above api make sense to you? Do you have suggestions for improvements, before I start coding?

RemiLehe:
As discussed in #137, this adds particle tracking in openPMD-viewer, and includes:
- the implementation of the `ParticleTracker` class
- allowing the user the pass either a `ParticleTracker` instance or a dictionary as `select`
- a new tutorial on particle selection and tracking.

When reviewing this pull request, I would advice to first have a look at the new tutorial. Note, for testing, that you will need to download the data **by hand** from the branch `add_particle_ids` of [openPMD-example-datasets](https://github.com/openPMD/openPMD-example-datasets), as the current `master` branch does not contain the required particle ids.
(For the same reason, the automatic tests will fail for this pull request, but you can still review it.)

A quick remark on the implementation: it requires to rapidly match the tracked particle ids with the existing ids at a given iteration. Much of this can be written in numpy (and is therefore fast) but there is one step that cannot, and which I had to write in explicit Python (i.e. using for loops).
~~Currently, this step is compiled with `numba` (otherwise the non-compiled version vastly dominates the time it takes to track particles), but in the future I would like to also have it implemented in Cython (for those users that do not have numba installed). Any volonteer is welcome :)~~
This step is now compiled using `Cython`, in order to avoid slow Python code.


opmd_viewer/openpmd_timeseries/particle_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+        pid = read_species_data(file_handle, species, 'id', extensions)
+        selected_indices = self.get_extraction_indices( pid )
+
+        # For each particle quantity, select only the tracked particles
+        for i in range(len(data_list)):
+            if len(data_list[i]) > 1:  # Do not apply selection on scalars
+                data_list[i] = self.extract_quantity(
+                    data_list[i], selected_indices )
+
+        return( data_list )
```
ax3l: note: this is [a bit superfluous](http://stackoverflow.com/a/4979486/2719194) for single argument returns, since `return` is not a function and nothing needs to be grouped.

It's a bit like writing `(a * b ) + c` so one should omit it.

opmd_viewer/openpmd_timeseries/data_reader/utilities.py
```diff
@@ -89,11 +92,13 @@ def get_data(dset, i_slice=None, pos_slice=None):
...
-    # Convert to float and scale by the conversion factor
-    if data.dtype != np.float64:
-        data = data.astype(np.float64)
-    if dset.attrs['unitSI'] != 1.0:
-        data *= dset.attrs['unitSI']
+    # Convert to the right type
+    if data.dtype != output_type:
+        data = data.astype( output_type )
+    # Scale by the conversion factor
+    if output_type is np.float64:
```
RemiLehe: Good point. I just changed this.

opmd_viewer/openpmd_timeseries/particle_tracker.py
RemiLehe: Yes, that's right, this is superfluous. However, it is used in many other places in the code. We can do a separate pull request to correct this throughout the code in a consistent manner.

RemiLehe:
@ax3l Regarding your remark on efficient select operations: I do use them in the current code:
```python
    def extract_quantity( self, q, selected_indices ):
        """
        Select the elements of the array `q`, so as to only return those
        that correspond to the tracked particles.
        ....
        """
        # Extract the selected elements
        selected_q = q[ selected_indices ]
```
The main point is that you can do this with an array of **array indices** (like `selected_indices`) but not with an array of **particle ids**, which is why I have to go through all the trouble of extracting the index that corresponds to each id (which in this case I had to do through a Python/Cython function)

Does this answer your question?

PR #141, opened 2017-02-15
Correct indexing of arrays with floats

RemiLehe:
The automated tests are currently failing because the most recent version of numpy does not allow indexing of arrays with floats (array indices have to be integers). In Python 3, because divisions return float by default, parts of the openPMD-viewer code was indexing arrays with floats.

This pull request fixes the issue.

I also updated the notebook number 5, because it was trying to access the iteration 200, which not part of example openPMD dataset anymore.


soerenjalas:
Thanks for the fix I'm still mostly using Python 2.7 so I wasn't aware of this.

RemiLehe:
OK, no worries. Thanks a lot for merging!

Issue #144, opened 2017-02-21
utilities `list_h5_files` relies on time step in file name


PrometheusPi:
Currently, all time series extractions are done using the `list_h5_files` function from `opmd_viewer.openpmd_timeseries.main`. This function extracts the time step from the last number before the suffix not from the time step in the hdf5 files. See code [here](https://github.com/openPMD/openPMD-viewer/blob/b7e84c060e7e049f532b324d8365a2b7c650b6cf/opmd_viewer/openpmd_timeseries/utilities.py#L47). Thus files like `libSplash` serial output will result in wrong time steps (the serial output extracts files with a given mpi rank as `filename_[timestep]_[rankx]_[ranky]_[rankz].h5`).

RemiLehe:
Thanks for pointing out this issue!
Would you have the time to do a pull request to correct this? Otherwise I can also do it.

PrometheusPi:
No problem. I will provide a pull request.

PR #145, opened 2017-02-21
extract time step from hdf5 directly 

PrometheusPi:
This pull request solves issue #144 by extracting the time step / iteration from hdf5 directly and not from the file name.

*However, extracting the time step from hdf5 files is slower.*

### DIFF
--- a/opmd_viewer/openpmd_timeseries/utilities.py
+++ b/opmd_viewer/openpmd_timeseries/utilities.py
@@ -3,14 +3,14 @@ This file is part of the openPMD-viewer.
 
 It defines a number of helper functions that are used in main.py
 
-Copyright 2015-2016, openPMD-viewer contributors
-Authors: Remi Lehe
+Copyright 2015-2017, openPMD-viewer contributors
+Authors: Remi Lehe, Richard Pausch
 License: 3-Clause-BSD-LBNL
 """
 
 import os
-import re
 import numpy as np
+import h5py
 from .data_reader.particle_reader import read_species_data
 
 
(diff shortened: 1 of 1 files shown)


opmd_viewer/openpmd_timeseries/utilities.py
```diff
@@ -38,17 +39,15 @@ def list_h5_files(path_to_dir):
...
-                iters_and_names.append((iteration, full_name))
+            full_name = os.path.join(
+                os.path.abspath(path_to_dir), filename)
+            # extract time step / iteration from first time
+            # entry in hdf5 file
+            f = h5py.File(full_name, 'r')
+            iteration = int(list(f['/data'].items())[0][0])
+            f.close()
+            # Create list of tuples (which can be sorted together)
+            iters_and_names.append((iteration, full_name))
```
ax3l: uh, append all found iterations, not only the first one :-)

ax3l:
cool, let's see if we can even open a `groupBased` file now: https://github.com/openPMD/openPMD-tools/issues/4

RemiLehe:
@PrometheusPi : Thanks, that was fast!
The automated tests indicate that there is a minor linting issue. Could you try running
```
pep8 --ignore=E201,E202,E122,E127,E128,E131 opmd_viewer
```
and fix the issues?

Otherwise, the PR is fine with me; thanks again!

PrometheusPi:
@RemiLehe Thanks for pointing out the pep8 issues - I think I fixed them. 
Now it fails [here](https://github.com/openPMD/openPMD-viewer/blob/b7e84c060e7e049f532b324d8365a2b7c650b6cf/tests/test_tutorials.py#L42) with a return value of 256.

RemiLehe:
The automated tests return the following error:
```
File "/home/travis/miniconda2/envs/testing/lib/python3.6/site-packages/openPMD_viewer-0.5.2-py3.6-linux-x86_64.egg/opmd_viewer/openpmd_timeseries/utilities.py", line 50, in list_h5_files

    iters_and_names.append((int(str_iteration), full_name))

TypeError: int() argument must be a string, a bytes-like object or a number, not 'Group'
```
My guess is that `.items()` should be replaced by `.keys()`. Could you check? 

Note: you can run the automated tests on your local computer by doing:
```
python setup.py install
python setup.py test
```

RemiLehe:
Also, I'm not sure if `[0]` is needed. On the whole, I guess the line
```python
iterations = list(f['/data'].items())[0]
```
should be replaced by
```python
iterations = list(f['/data'].keys())
```
But again, I have not checked.

PrometheusPi:
@RemiLehe Thanks! 
Using `.keys()` is even simpler and faster - thanks for the input.

RemiLehe:
Great, I'll merge. Thanks for the PR.

Of course, if you need this feature in production, you'll need to install the `dev` version, at least until our next release.

PrometheusPi:
@ax3l I tested the `groupBased` file and it works like a charm :+1:

ax3l:
awesome, thanks! :dancing_men:

Issue #64, opened 2016-04-20
OpenPMDTimeSeries: Unit Systems


ax3l:
To improve the generality of the viewer without harming its usefulness for a specific domain, we could make the following adjustment regarding unit systems:

Per default, the `OpenPMDTimeSeries` should not convert, rename or exclude records.
But we could set a unit system that is used for reading and formatting plots in a way such as:

```python
from opmd_viewer import OpenPMDTimeSeries
import opmd_viewer.unit_systems

ts = OpenPMDTimeSeries('...')
# change to lambda0 & c based system
ts.set_unitsystem(unit_systems.LPA(800.e-9))
# change back to SI
ts.set_unitsystem(unit_systems.SI())
# change a plasma system without a defined laser
ts.set_unitsystem(unit_systems.Plasma(1.e15))

# read or plot data now
```

### Example Unit Systems (Pre-Defined)

- `SI()`: just forward what we read scaled by `unitSI`
- `LPA(lambda0)`: scale length to `lambda0`, times to `lambda0/c`, speed to `c`, ...
- `Plasma(omega_pe)`: scale time by `(2 pi) / omega_pe`, speed by `c`, length by `(c * 2 pi) / omega_pe`, ...
- ...

### Re-Naming of Records

During the change of the unit system we can also rename the records since e.g., a momentum in beta gamma (`ux`) is more readable and known in a specific domain.

### Slider

The `slider()` could add an optional `exclude_particle_records=list` argument which can be used to exclude `charge` and `mass` as its currently hard-coded.

ax3l:
This change would go great with the change of "some calls return microns and fs" since one could switch to a unit-system with the release update that still keeps that convention.

ax3l:
Just found that *yt* has a nice (planned) way to handle units, also taking care of the angular measure but ignoring two base units (mol & cd): [YTEP-0028](https://ytep.readthedocs.io/en/latest/YTEPs/YTEP-0028.html)

pgrete:
This has been open for quite some time now.
Have people been using work around or are there plan to make this happen?
We're currently wondering how to make this transparent for downstream users of simulation data (which, in the astro community, mostly comes aligned with cgs units)?

simbot912:
I've been reading through this and the codebase, and would like to help. Since the full framework (LPA/Plasma scaling + record renaming) is a large surface, I'd suggest landing it in stages so each PR stays small and reviewable:

**Stage 1 — foundation (small, low-risk):**
[...]

PR #157, opened 2017-04-20
Close #156 Need normalization for zero-mass particles

### LINKED ISSUE #156: Need normalization for zero-mass particles
With the introduction of bremsstrahlung and Compton scattering into PIConGPU a photon model has been added to the available particle species. Since a photon's mass is zero we encounter a divide by zero during the normalization of the momentum data to (mc)^-1. 
```
/lib/python3.4/site-packages/opmd_viewer/openpmd_timeseries/data_reader/particle_reader.py:84: 
RuntimeWarning: divide by zero encountered in true_divide
  norm_factor = 1. / (get_data(species_grp['mass']) * constants.c)
```
In this case it would be desirable to have an arbitrary normalization and maybe a logarithmic plot scaling as a default. Then the users could set a normalization value by themselves after seeing the logarithmic plot and deciding where the regions of interest are and maybe change back to linear scaling. :grinning:

RemiLehe:
Close #156: This pull request avoids dividing the momentum by m*c in the case of particles having zero mass. Thus, for these particles, the momentum is effectively kept in SI units (kg.m.s^-1).

@n01r : Does this work for your use case? If not, please let me know how it could be improved.

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/particle_reader.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/particle_reader.py
@@ -81,8 +81,11 @@ def read_species_data(file_handle, species, record_comp, extensions):
         data *= 1.e6
     # - Return momentum in normalized units
     elif record_comp in ['ux', 'uy', 'uz' ]:
-        norm_factor = 1. / (get_data(species_grp['mass']) * constants.c)
-        data *= norm_factor
+        m = get_data(species_grp['mass'])
+        # Normalize only if the particle mass is non-zero
+        if np.all( m != 0 ):
+            norm_factor = 1. / (m * constants.c)
+            data *= norm_factor
 
     # Return the data
     return(data)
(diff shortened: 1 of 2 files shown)


ax3l:
thanks for the quick PR!

Admittedly that's a bit of an odd unit for photons (as long as no-one is doing gluon or graviton science with it, where it's also odd). I guess the default range is still [0.:1] for kg m/s or can we make a more educated initial guess? Normalizing massless particles to electron quantities might not be too bad after all (eV/keV/MeV)?

n01r:
Thanks!

Well, in the histogram plots the position is displayed in microns by default. So a microscopic setup is assumed. Photons in these setups would most likely originate from collective electronic processes, atomic processes and nuclear processes - or being emitted when charged particles interact with strong e.m. fields. From that point of view I'd consider electron volts to be most intuitive.

ax3l:
Intuitively, I am ok with the PR as it is right now (SI) as it is actually a job of issue #64 to provide a meaningful unit system for a domain

RemiLehe:
I agree that SI might not be ideal for photons. But at least it is now possible to extract the data (instead of getting an Exception), and thus the user can rescale it to whatever units they deem fit :)

PR #159, opened 2017-05-15
Fix the releasing process for conda

RemiLehe:
When creating a new release of `openPMD-viewer`, we upload the package so that it can be installed with `pip`, as well as with `conda`.

In the case of `conda`, this requires `conda build`. However, recently I experienced issues with `conda build`ing openPMD-viewer on Mac OSX (which is the OS on which I typically create the package). As a consequence, this pull request provides tools to create the openPMD-viewer conda package **inside a Docker container**. (The Docker container is a virtual linux environment, and thus it does not have the above mentioned problem of Mac.)

More precisely:
- I modified RELEASING.md which gives instructions on how to build the package.
- I replaced `build.sh` (which automated the process of creating openPMD-viewer packages for python 2.7, 3.4, 3.5, 3.6) by a corresponding Dockerfile.

### DIFF
--- a/conda_recipe/Dockerfile
+++ b/conda_recipe/Dockerfile
@@ -1,0 +1,19 @@
+FROM continuumio/miniconda
+
+RUN apt-get update && apt-get install -y gcc libgl1-mesa-glx
+
+RUN conda install --yes conda conda-build
+
+CMD cd /home/ \
+    && conda build --python=2.7 . \
+    && conda build --python=3.4 . \
+    && conda build --python=3.5 . \
+    && conda build --python=3.6 . \
+    && conda convert $(conda build --python=2.7 . --output) -p osx-64 \
+    && conda convert $(conda build --python=2.7 . --output) -p linux-64 \
+    && conda convert $(conda build --python=3.4 . --output) -f -p osx-64 \
+    && conda convert $(conda build --python=3.4 . --output) -f -p linux-64 \
+    && conda convert $(conda build --python=3.5 . --output) -f -p osx-64 \
+    && conda convert $(conda build --python=3.5 . --output) -f -p linux-64 \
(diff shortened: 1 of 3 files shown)


ax3l:
building it in a clean and controlled environment is a splendid approach, thank you for the update!

conda_recipe/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM continuumio/miniconda
+
```
ax3l: usually you add a maintainer filed for a dockerfile, which would be you

conda_recipe/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN apt-get update && apt-get install -y gcc libgl1-mesa-glx
```
ax3l: to keep layers small one adds a `&& rm -rf /var/lib/apt/lists/*` at the end of a apt-get line in a container

(this also has a security plus for people using your container to be forced taking new repos in first via `update` but that's less important here)

RemiLehe:
Thanks for your prompt review! I made the changes that you suggested. As you foresaw, the option `--no-install-recommends` caused the build to fail, so I did not include it. 

You can merge if you agree with the changes.

conda_recipe/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM continuumio/miniconda
+MAINTAINER Remi Lehe <rlehe@lbl.gov>
+
+RUN apt-get update \
+    && apt-get install -y \
+        gcc \
+        libgl1-mesa-glx \
```
ax3l: @RemiLehe why do you need `libgl1-mesa-glx`? Builds without it for me.

ax3l: ah no, you are right `PyQt5` needs it when running the CMD

Issue #140, opened 2017-02-10
Keep zoom level when changing time step


HighIander:
Currently, the zoom level restores to unzoomed when the time step is changed. This might be unconvenient when one wants to only see a detail and study its change over time.
Keeping the zoom level, just as all the other settings are kept would be very helpful.

Off-topic: Cool tool!!!!

ax3l:
Sounds great! We have to remember it for both plots in the `slider` (field plot and histograms) I guess, making it a bit complex. Also we are currently lagging a bit the `mpl` figure handling, which leads in `%matplotlib notebook` node often to the case that later matplotlib calls interfere with the slider (independent issue).

I think the required basis to get this done is what we discussed briefly when I was last time at LBNL about a [MVC](https://en.wikipedia.org/wiki/Model-view-controller) (model-view-controller) abstraction of the slider, especially which members (figures, axes, etc.) need to be kept as members.

I did something similar last year with a small quasi-GUI tool where I kept the following members:
```python
        self.fig #1 & 2
        self.ax # 1 & 2
        self.cb # 1 & 2
        self.last_xlim = None # 1 & 2
        self.last_ylim = None # 1 & 2
        self.myCmaps = {} # 1 & 2
```

Did changes in a `draw_iteration(...)` call which only updated existing `axes` (or created them initially) and ended in
```python
self.fig.canvas.draw()
```
Members where updated with e.g. zoom information via
```python
    def save_zoom(self) :
        self.last_xlim = self.ax.get_xlim()
        self.last_ylim = self.ax.get_ylim()

    def reset_zoom(self) :
        self.last_xlim = None
        self.last_ylim = None
```
before starting a new draw (in case on can not react on an event).

[The following matplotlib docs](https://gist.github.com/ax3l/fc123cb94f59d440f952) helped me tremendously understanding all the figure vs axes vs axis vs formatter vs canvas vs ... and redraw fuzz.

RemiLehe:
Note: This issue is partially addressed by PR #165 and #167

ax3l:
thank you so much! ✨

PR #165, opened 2017-07-18
Set plot range for field plotter

RemiLehe:
This pull request is the first of a series of 3 pull request that (in my opinion at least) improve the openPMD-viewer GUI.

The current pull request adds the possibility for the user to enter a range of values, in order to clip the plot of the fields. Here is how it looks
screen shot 2017-07-18 at 7 53 51 am

If "Use this range" is not used, then the full range of the box (along this axis) is plotted. 

Note: For now, this does not implement the solution proposed in #140. This is in part because, when using `%matplotlib inline`, the user is unable to interactively zoom, and therefore needs a separate widget.

### DIFF
--- a/opmd_viewer/openpmd_timeseries/plotter.py
+++ b/opmd_viewer/openpmd_timeseries/plotter.py
@@ -131,7 +131,7 @@ class Plotter(object):
         plt.title("%s:   t =  %.1f fs   (iteration %d)"
                   % (species, time_fs, iteration), fontsize=self.fontsize)
 
-    def show_field_1d( self, F, info, field_label, current_i,
+    def show_field_1d( self, F, info, field_label, current_i, plot_range,
                             vmin=None, vmax=None, **kw ):
         """
         Plot the given field in 1D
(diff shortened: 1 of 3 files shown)


ax3l:
@RemiLehe pls rebase against `dev` :)

opmd_viewer/openpmd_timeseries/plotter.py
```diff
@@ -195,6 +204,10 @@ def show_field_2d(self, F, info, slicing_dir, m,
 
         geometry: string
            Either "2dcartesian", "3dcartesian" or "thetaMode"
+
+        plot_range : list of lists
```
ax3l: a list of a `.min` & `.max` dict (tuple) would probably be easier to read but ok :)

RemiLehe: Yeah, I agree there could be different options here. I was following the same convention as [matplotlib's hist2d](https://matplotlib.org/devdocs/api/_as_gen/matplotlib.axes.Axes.hist2d.html).

Since this convention is also used by the other upcoming pull requests, would you be okay with keeping your remark in mind, and then maybe dealing with it, in bulk, once the 3 upcoming pull requests are merged?

If you are

PR #166, opened 2017-07-18
Better color widgets

RemiLehe:
**Wait for PR #165 to be merged before reviewing**

The widget for the colorbar (with the double slider to pick min/max values) was quite cumbersome in practice (in part because it was small and imprecise). I replaced it by a set of text widget, which look like this:

screen shot 2017-07-18 at 8 12 09 am

Also, the code for the colorbar widget has been refactored and is now more concise and cleaner.

### DIFF
--- a/opmd_viewer/openpmd_timeseries/interactive.py
+++ b/opmd_viewer/openpmd_timeseries/interactive.py
@@ -123,7 +123,7 @@ class InteractiveViewer(object):
                     clear_output()
 
                 # Colorscale range
-                vmin, vmax = fld_color_button.get_range()
+                vmin, vmax = ptcl_color_button.get_range()
 
                 if ptcl_yaxis_button.value == 'None':
                     # 1D histogram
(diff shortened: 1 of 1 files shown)


RemiLehe:
FYI: I rebased the branch on `dev`.

opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -500,6 +452,82 @@ def convert_to_int(m):
...
+    Class that allows to select a colorbar and the corresponding range.
+    It features a widget for the exponent, in order to rapidly change
+    the order of magnitude of the colorbar.
+    """
+
+    def __init__( self, callback_function, default_cmap ):
+        """
+        Initialize a set of widgets that select a colorbar.
+
+        # TODO
```
ax3l: here is a todo :)

RemiLehe: Good catch! I forgot to add a documentation. I'll update the PR.

PR #167, opened 2017-07-18
Set plot range for the particle plotter

RemiLehe:
**Please merge #166 first before reviewing this pull request**

This PR is similar to PR #165. This allows the user to optionally impose the range between which the particle histograms are being performed. This particularly useful e.g. when plotting the phase space in LWFA, since in this case the longitudinal phase space is extremely elongated (along the momentum axis), and having a good binning resolution would require a very large number of bins if the user were to plot the **full range**.

Here is how it looks:
screen shot 2017-07-18 at 8 03 47 am
If "Use this range" is not checked for a given direction, then openPMD-viewer check whether "Use field mesh" is checked. If yes it will use the size of the field mesh. Otherwise it will just determine the range from the min/max of the particle arrays.


RemiLehe:
Rebased.

PR #168, opened 2017-09-06
Correct widget dimensions for compatibility with ipywidgets version 4, 5 and 6

RemiLehe:
Because ipywidgets version 4, 5 and 6 layout the widgets in different manners, I had to tweak the position and size of widgets in the "Plotting options" section, so that they fit inside the panels with all the versions. 

In particular, this involved removing the label "Use this range", as well as putting the widget "x10^" on a new line for ipywidgets 4. 

Here is how the widgets look with:

- ipywidgets 6 (quite neat)
ipywidgets_6

- ipywidgets 5
ipywidgets_5

- ipywidgets 4 (the extra spacing between widgets is unwanted, but cannot be fixed easily: https://github.com/jupyter-widgets/ipywidgets/issues/709)
ipywidgets_4

### DIFF
--- a/opmd_viewer/openpmd_timeseries/interactive.py
+++ b/opmd_viewer/openpmd_timeseries/interactive.py
@@ -509,33 +509,21 @@ class ColorBarSelector(object):
         """
         # Set the widget dimensions
         set_widget_dimensions( self.active, width=20 )
-        if ipywidgets_version >= 5:
-            text_width=30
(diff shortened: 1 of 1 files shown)


RemiLehe:
Note: once this is merged, I will create a new release of openPMD-viewer.

soerenjalas:
Thanks for the fix!

PR #169, opened 2017-10-09
Prevent failure of emittance calculation for empty arrays

RemiLehe:
The function `get_emittance` raises an error when the array of particles is empty. This does not occur with `get_divergence` for instance, because of some specific safeguards in the function `wstd`. 

This pull request adds similar safeguards for the calculations of averages in the function `get_emittance`. 

Note: I named the new averaging function `w_ave` (instead of `wave` which might be confusing). For consistency, I also renamed `wstd` as `w_std`.


soerenjalas:
Thanks for the fix! As the renaming of wstd to w_std is technically an API change we should warn about this in the next release notes. I don't expect that anybody other than me has used this method outside of this class, but anyways…

RemiLehe:
Thanks for letting me know. I did not realize someone was actually using the `wstd` function. I'll add this in the `CHANGELOG.md` when doing a new release.

PR #170, opened 2017-10-11
Adapt envelope calculation for 2D arrays

RemiLehe:
@dpgrote recently pointed out to me that the result of `get_laser_envelope` sometimes looked weird in 2D, with sharp horizontal lines, as in the example below:
cut

Upon further investigation, he found that this is because the calculation of the central frequency (around which filtering is performed in order to get the envelope) is performed independently for each horizontal line. Thus, each horizontal line is potentially filtered differently.

This pull request solves this problem by calculating the central frequency only once, for the whole 2D array. This is done by generalizing the function `_fft_filter` so that it can accept both 1D arrays and 2D arrays as input (instead of only 1D arrays before).

Here is the same image with the code from this PR:
uncut

### DIFF
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -430,17 +430,16 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
                                 coord=pol, theta=theta, m=m,
                                 slicing_dir=slicing_dir )
         info = field[1]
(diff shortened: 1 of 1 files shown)


RemiLehe:
By the way, many thanks to @dpgrote for identifying and reporting this issue!!

soerenjalas:
I don't use the 2d envelope a lot. Thanks for pointing out and fixing the problem!

PR #171, opened 2017-10-23
Close #154: Plot kwargs are not used

### LINKED ISSUE #154: Plot kwargs are not used
When using the slider of `openpmd-viewer` one can pass keyword arguments but they will not be used.

RemiLehe:
This pull request modifies the slider code so that it uses the user-provided kwargs.

As explained in the docstring, the kwargs will be applied both to the particle plots and field plots, and set the initial plotting options. The user can then still modify these options through the slider interface.


ax3l:
closes #154

PR #172, opened 2017-10-24
Allow openPMD-viewer to work without matplotlib

RemiLehe:
In some use case, openPMD-viewer can be used solely to read openPMD files and convert them into numpy arrays. In these cases, matplotlib is not needed, and it can be a significant overhead to install/include matplotlib. (One example is when using openPMD-viewer to read checkpoint files with FBPIC, or using openPMD-viewer to read a beam distribution when doing external injection with Warp)

For this reason, it would be advantageous if matplotlib is not a strict dependency of openPMD-viewer. This is what this pull request implements. When matplotlib is not present, the user will receive an exception whenever trying to plot something with openPMD-viewer ; but otherwise the code will work fine.


opmd_viewer/openpmd_timeseries/plotter/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+# and show_field_2d. If matplotlib is not installed, an exception will be
+# raised when these methods are called.
+
+try:
+    # Check wether the plotter can be loaded
+    from .plotter import Plotter
+except ImportError:
+    # Otherwise, create a dummy class which returns an error message
+    # when the method `slider` is called.
+    class Plotter(object):
```
ax3l: I would also put this try-except and fallback class in `plotter.py`

requirements.txt
```diff
@@ -1,4 +1,3 @@
 numpy
 scipy
-matplotlib
```
ax3l: as with Jupyter, it would make sense to define sub-packages/requirements. otherwise we will loose track what the full requirements are, e.g. to also have the widged GUI and the plotting via MPL.

You can for example add [`extras`](http://peak.telecommunity.com/DevCenter/setuptools#declaring-extras-optional-features-with-their-own-dependencies) such as `GUI` and `plot` to `setup.py`:
```python
setup(
    # [...]
    extras_require = {
        'GUI':  ["ipywidgets>=A.B"],
        'plot': ["matplotlib>=X.Y"],
    }
)
```
([Newer doc](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras))

We can also define an (pip install) `openPMD-viewer[all]` install target that will call to install the GUI and the plot (MPL) deps as well for proper packaging. See [the logic in IPython's setup](https://github.com/ipython/ipython/blob/dae349b90e0fe3ad37c08792f33d4b2d35dabc7c/setup.py#L172-L232) for example.

Or in a more manual case just:
```python
setup(
    # [...]
    extras_require = {
        'GUI':  ["ipywidgets>=A.B"],
        'plot': ["matplotlib>=X.Y"],
        'all': ["ipywidgets>=A.B", "matplotlib>=X.Y"],
    }
)
```

RemiLehe:
Thanks for the great suggestions! I'll modify the PR accordingly.

RemiLehe:
Ready for second round of review!

PR #173, opened 2017-10-24
Ipywidgets error

RemiLehe:
Currently openPMD-viewer prints an (ugly-looking) message whenever ipywidgets is not installed. This is a bit superfluous, since openPMD-viewer can be used for scripting or other non-jupyter uses, and since in this case, the error message is unneeded.

This pull request modifies the code so that an error message is printed only if the user attempts to call the `slider` method.

### DIFF
--- a/opmd_viewer/openpmd_timeseries/interactive.py
+++ b/opmd_viewer/openpmd_timeseries/interactive.py
@@ -8,14 +8,17 @@ Copyright 2015-2016, openPMD-viewer contributors
 Authors: Remi Lehe, Axel Huebl
 License: 3-Clause-BSD-LBNL
 """
-
-from ipywidgets import widgets, __version__
-from IPython.core.display import display, clear_output
 import math
-import matplotlib
-import matplotlib.pyplot as plt
 from functools import partial
-ipywidgets_version = int(__version__[0])
+try:
+    from ipywidgets import widgets, __version__
+    ipywidgets_version = int(__version__[0])
+    from IPython.core.display import display, clear_output
+    import matplotlib
+    import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


opmd_viewer/openpmd_timeseries/interactive/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# This parent class defines the method `slider` ; if the dependencies are
+# unavailable, an error message will be returned.
+
+try:
+    # Check wether the interactive interface can be loaded
+    from .interactive import InteractiveViewer
+except ImportError:
+    # Otherwise, create a dummy class which returns an error message
+    # when the method `slider` is called.
+    class InteractiveViewer(object):
```
ax3l: that an ok-ish work-around but one could keep the implementation of the "fallback" `InteractiveViewer` class in `interactive.py` and implement the `try-except` case there. this avoids taking the signature of the class out to the places where it is used

RemiLehe:
Good point! I'll modify the PR.

RemiLehe:
Ready for review again.

opmd_viewer/openpmd_timeseries/interactive.py
```diff
@@ -40,6 +43,11 @@ def slider(self, figsize=(6, 5),
         kw: dict
             Extra arguments to pass to matplotlib's imshow
         """
+        # Check that the dependencies have been installed
+        if not dependencies_installed:
```
ax3l: instead of defining a global one could also just try to access, e.g. `widgets` here and re-through the own error if it fails.

ax3l:
ah actually it's fine, imho :)

Issue #175, opened 2017-11-06
invisible 'active' checkbox


hightower8083:
Hi Remi

I've got an invisible 'active' checkbox in Firefox (in a fresh installation of Debian stretch with conda3)
The bug appears to be sensitive to two things
- the fact that checkbox is included into widgets.HBox
- the width fixed

I've got it fix by doing the following modif (for example in to_container of ColorBarSelector class)

```python
        # Set the widget dimensions
#        set_widget_dimensions( self.active, width=20 )
        set_widget_dimensions( self.low_bound, width=60 )
        set_widget_dimensions( self.up_bound, width=60 )
        set_widget_dimensions( self.exponent, width=45 )
        set_widget_dimensions( self.cmap, width=200 )
        # Gather the different widgets on two lines 
        cmap_container = widgets.HBox( children=[  
            widgets.HTML( "<b>Colorbar:</b>"), self.cmap ])
        if ipywidgets_version > 4:
            # For newer version of ipywidgets: add the "x10^" on same line
            range_container = widgets.HBox( children=[             
                add_description("from", self.low_bound, width=30 ),
                add_description("to", self.up_bound, width=20 ),  
                add_description("x 10^", self.exponent, width=45 ) ] )
            final_container = widgets.VBox( 
                children=[ cmap_container, range_container, active_container, self.active ])
```

in my case, such was needed for all other to_container methods.
The result looks like this -- is not too ugly, so if you wish i could PR this or you may find some elegant solution

![opmd_ntbk_bugfix](https://user-images.githubusercontent.com/22051570/32431091-ae05f714-c2da-11e7-835e-4c4584f04339.png)

RemiLehe:
Hey, thanks a lot for reporting this issue!
Could you let me know which version of `jupyter` and `ipywidgets` you have? I'll try to reproduce the issue.

Commenting out the `set_widget_dimensions( self.active, width=20 )` could be solution, but it would  maybe be problematic for other version of `ipywidgets` (for some versions, the layout of the widgets is quite different, and this is why I impose a set width.)

hightower8083:
right, forgot to type the versions:
`ipython` 6.1.0
`ipywidgets` 7.0.0
`jupyter` 1.0.0 (client 5.1.0)
`mozilla firefox ESR` 52.4.0 (doubt its relevant)

Its true that the way `pywidgets` develops is rather messy -- i've also had few widgets-based little apps which started to behave weird after upgrades..

[...]

Issue #174, opened 2017-10-24
Slider vanishing with Jupyter 5.2.0


n01r:
I recently upgraded my Jupyter to kernel version 5.2.0 to use some nifty `nbextensions`. When I tried to use the openPMD slider again, at first, everything was shown normally but as soon as I picked a time step only the plots were shown but the slider widget had vanished.

I'm using Python 3.4.3 and Jupyter 5.2.0.

RemiLehe:
Thanks for reporting this! Could you check which ipywidgets version you have?

n01r:
Sure! I have version 7.0.2.

RemiLehe:
OK. Could you try typing, in a terminal: 
```
jupyter nbextension enable --py widgetsnbextension
```
and then restarting your notebook?

n01r:
Didn't work ^^; same behavior

RemiLehe:
Yes, sorry I misunderstood your issue. I can now reproduce the issue that you have.
I'll let you know once I find a solution.

In the meanwhile, I think that downgrading to ipywidgets 6 would fix your issue.

RemiLehe:
Or another option is to use `%matplotlib` or `%matplotlib notebook` intead of `%matplotlib inline`

n01r:
Thanks! Well, using `%matplotlib notebook` already fixes it so the solution to this it not time-critical. :+1:

RemiLehe:
The new version of openPMD-viewer will now issue a warning in this configuration (see PR #176).

To answer one of @ax3l's questions: the disappearance of the `slider` is explained by the [release note of ipywidgets 7](https://ipywidgets.readthedocs.io/en/stable/changelog.html):
```
Widgets are now displayed in the output area in the classic notebook and are treated as any other output. This allows the widgets to work more naturally with other cell output. To delete a widget, clear the output from the cell. Output from functions triggered by a widget view is appended to the output area that contains the widget view. This means that printed text will be appended to the output, and calling clear_output() will delete the entire output, including the widget view.
```
In the case of `openPMD-viewer`, when the user had `%matplotlib inline`, the plots are appended to the output area (resulting, by default, in more and more plots stacking in the output area, as the user uses the slider). In order to avoid this stacking, `openPMD-viewer` calls `clear_output` in this case. With previous versions of `ipywidgets`, this only cleared the plots (to make room for new plots), but now with `ipywidgets 7` the slider also gets cleared.

[...]

PR #176, opened 2017-11-07
Support for ipywidget 7

RemiLehe:
This pull request modifies the slider so as to have better behavior with ipywidgets 7.

In particular:
- A user warning is issued when `ipywidgets 7` is used in conjunction with `%matplotlib inline` (see issue #174).
- The size of the `ToggleButtons` which contain the name of fields and particle quantities is adjusted to the length of the name (this feature was removed with `ipywidgets 6`, but reintroduced in `ipywidgets 7` if the code specifically asks for it)
- The `CheckBox` is corrected so that it displays correctly (see issue #175)

Here is how the `slider` looks like with `ipywidgets 7`:
screen shot 2017-11-07 at 2 48 21 pm

### DIFF
--- a/opmd_viewer/openpmd_timeseries/interactive.py
+++ b/opmd_viewer/openpmd_timeseries/interactive.py
@@ -667,7 +667,7 @@ class ParticleSelectWidget(object):
         self.n_rules = n_rules
 
         # Create widgets that determines whether the rule is used
-        self.active = [widgets.Checkbox(value=False)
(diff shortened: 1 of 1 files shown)


soerenjalas:
Thanks for the PR. I fine with merging if Axel has no objections.

ax3l:
I will still merge this for now, my questions can be addressed in a follow-up PR if needed

PR #179, opened 2017-11-09
Remove matplotlib dependency from LpaDiagnostics addon

### LINKED ISSUE #178: LpaDiagnostic still depends on matplotlib
While PR #172 removed the dependency on `matplotlib`, the `LpaDiagnostic` still depends on `matplotlib`. This should be corrected by using the same technique as in PR #172

soerenjalas:
Close #178. On import it is now checked if mpl is installed. If plots are requested without matplotlib being installed a runtime error is thrown just like in the plotter class. 

I hope its fine that my branch is based on the current master, if not I will redo the changes with the dev branch.


RemiLehe:
Great! Thanks a lot for this PR!
Actually, it is good that you are doing from master, since apparently some previous commits were missing on `dev`.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -281,12 +287,12 @@ def get_emittance( self, t=None, iteration=None, species=None,
         # Calculate the necessary RMS values
         x *= 1.e-6
         y *= 1.e-6
-        xsq = np.average( x ** 2, weights=w )
-        ysq = np.average( y ** 2, weights=w )
-        uxsq = np.average( ux ** 2, weights=w )
-        uysq = np.average( uy ** 2, weights=w )
-        xux = np.average( x * ux, weights=w )
-        yuy = np.average( y * uy, weights=w )
+        xsq = w_ave( x ** 2, weights=w )
```
ax3l: note: usually, ppl shorten average with `avg` ;)

PR #181, opened 2017-11-21
Generalize the code to handle files containing different geometries

RemiLehe:
@caozigao recently pointed out that the `openPMD-viewer` does not work properly when a file contains different geometries (e.g. 2d fields and 3d fields). (Btw, thanks for reporting the issue!)

This pull request generalizes the code so as to fix this issue. I tested the code with the example created by [this script](https://github.com/openPMD/openPMD-validator/blob/fix_example/createExamples_h5.py), which contains both cylindrical and 2d fields.

Here is a summary of the changes in this PR:
- The attributes `self.geometry` and `self.available_circ_modes`, etc. have been replaced by `self.fields_metadata` which contains this information **for each field**. This is then used in `get_field`, retrieving the relevant metadata **for the field that the user requests**.
- The attribute `self.avail_fields` used to be a dictionary, showing whether a given field is a vector or a scalar. This information is now moved to `self.fields_metadata`, and instead `self.avail_fields` is simply a list of the available fields.
- I also added an attribute `self.avail_geom` which lists all geometries in the present file. This convenient when creating the slider.
- One difficulty was the anti-aliasing code (see PR #132). There, to avoid aliasing when doing particle histograms, we were using information of the grid to choose the histogram bin size. However, when different geometries are used in the same file, it is unclear which grid information to use. To keep the code simple, I made the choice of picking the grid with the highest dimension (e.g. 3d rather than 2d, when available).
- I also updated the slider, so that the relevant buttons appear for files that contain different geometries. Some of the buttons even activate/deactivate depending on the geometry of the requested field (but not all, though: for instance the radial field component `r` is still available for cartesian fields, if the file also contains radial fields... This could be fixed in a future PR.)

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/field_reader.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/field_reader.py
@@ -329,9 +329,9 @@ def get_grid_parameters( dfile, avail_fields, metadata ):
     # (This function is for the purpose of histogramming the particles;
     # in this case, the highest dimensionality ensures that more particle
     # quantities can be properly histogrammed.)
-    geometry_ranking = {'1dcartesian':0, '2dcartesian':1,
-                        'thetaMode':2, '3dcartesian':3}
-    fields_ranking = [ geometry_ranking[ metadata[field]['geometry'] ] \
+    geometry_ranking = {'1dcartesian': 0, '2dcartesian': 1,
+                        'thetaMode': 2, '3dcartesian': 3}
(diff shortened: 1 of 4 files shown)


ax3l:
That's a very good aspect and similar to a question we had before about mixed mesh records in the same file: https://github.com/openPMD/openPMD-standard/issues/128

The standard does not forbid those and until know we left it open to data readers if they support multiple layouts (shapes & geometries). This PR is super useful!

PR #182, opened 2017-12-02
Tutorials: Binder

ax3l:
Enables [Binder](https://mybinder.org/v2/gh/ax3l/openPMD-viewer/topic-binder?filepath=tutorials%2F) support for our tutorials.

With binder, our users can quickly expore the features of openPMD-viewer online - without the need to install anything. It's just amazing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![Build Status master](https://img.shields.io/travis/openPMD/openPMD-viewer/master.svg?label=master)](https://travis-ci.org/openPMD/openPMD-viewer/branches)
 [![Build Status dev](https://img.shields.io/travis/openPMD/openPMD-viewer/dev.svg?label=dev)](https://travis-ci.org/openPMD/openPMD-viewer/branches)
 [![pypi version](https://img.shields.io/pypi/v/openPMD-viewer.svg)](https://pypi.python.org/pypi/openPMD-viewer)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -23,9 +24,12 @@ visualize the data.
 
 ### Tutorials
 
-The notebooks in the folder `tutorials/` demonstrate how to use both
-the API and the interactive GUI. You can view these notebooks online
-[here](https://github.com/openPMD/openPMD-viewer/tree/master/tutorials).
```
RemiLehe: I would still keep the paragraph
```
The notebooks in the folder `tutorials/` demonstrate how to use both
the API and the interactive GUI. You can view these notebooks online
[here](https://github.com/openPMD/openPMD-viewer/tree/master/tutorials).
``` 
and only add the paragraph
```
Alternatively you can even [run this tutorials online](https://mybinder.org/v2/gh/openPMD/openPMD-viewer/master?filepath=tutorials%2F)!
```
The main reason is that, in my experience, Binder is often down, (see e.g. [here](https://github.com/binder-project/binder/issues/95)) and so the user should have an easy way to still see the tutorials if this happens.

ax3l: good point, updated the PR accordingly.

RemiLehe:
Thanks for updating the PR!

PR #183, opened 2017-12-02
Slider figure

RemiLehe:
When using `%matplotlib figure`, it is often needed to explicitly pass the figure number through the slider (otherwise the figure that gets updated is often at a completely different position in the notebook). 

However, when recreating/updating a slider, rerunning the cell that created it resets the figure number. This pull request allows the user to pass the figure number directly in the `slider` call instead (see the tutorial example).

I also updated the tutorial, so that `%matplotlib notebook` is used. This is because `%matplotlib inline` does not work well with the latest version of `ipywidgets` (see https://github.com/openPMD/openPMD-viewer/pull/176)

### DIFF
--- a/tutorials/3_Introduction-to-the-GUI.ipynb
+++ b/tutorials/3_Introduction-to-the-GUI.ipynb
@@ -19,10 +19,8 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
-   "metadata": {
-    "collapsed": false
-   },
+   "execution_count": null,
+   "metadata": {},
    "outputs": [],
    "source": [
     "import os, sys, tarfile, wget\n",
@@ -51,13 +49,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
-   "metadata": {
-    "collapsed": false
-   },
+   "execution_count": null,
+   "metadata": {},
    "outputs": [],
    "source": [
-    "%matplotlib inline"
+    "%matplotlib notebook"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


ax3l:
Ah that is fantastic, I had this issue with follow-up plots for quite some time. Good you also change the default mpl in the tutorials.

tutorials/3_Introduction-to-the-GUI.ipynb
```diff
@@ -51,13 +49,11 @@
...
-   "execution_count": 2,
-   "metadata": {
-    "collapsed": false
-   },
+   "execution_count": null,
+   "metadata": {},
    "outputs": [],
    "source": [
-    "%matplotlib inline"
+    "%matplotlib notebook"
```
ax3l: @RemiLehe can you update this in all `.ipynb` files inside `tutorials/` please for consistency?

RemiLehe: I agree this would be more consistent, but it would complicate the other notebooks I think.

The reason is that many of the tutorials feature plots, which are shown just below the command that generated them. If I used `%matplotlib notebook`, this would no longer be the case, unless I add a `plt.figure(<new_number>)` before each command. However, I feel that this would make the notebook less readable.

Thus, would you be okay with keeping `inline` in these other notebooks.

ax3l: I understand, but wouldn't explicit figures instead of implicit figure assignment make sense in general? I see many people using the slider or some plots from the other APIs and own-read data together in the same notebook and using `%matplotlib notebook` as the default. This often leads to very confusing re-plots in figures they did not expect (because we don't show them the concept of figures).

PR #185, opened 2018-01-04
Implement Cloud-In-Cell deposition (i.e. linear weights) for histogramming

RemiLehe:
In `openPMD-viewer`, the particles are plotted through histograms. However, when particles are evenly-spaced, the histograms can have aliasing effects if the bin size is not a multiple of the spacing between particles. Currently, `openPMD-viewer` avoids this effect by choosing a bin size which is a multiple of the grid spacing. (see [PR #132](https://github.com/openPMD/openPMD-viewer/pull/132))

However, the information about grid spacing is not always available (e.g. if the fields data is not written in the openPMD file). In addition, the spacing between particles does not always align with the grid. 

For these reasons, this PR implements Cloud-In-Cell histogramming (as opposed to Nearest-Grid-Point histogramming), since this strongly reduces the aliasing effect, even when the grid spacing is not "matched" to the particle spacing. 

In practice, this PR adds the option `histogram_deposition` for the `get_particle` method. If `histogram_deposition` is set to `ngp`, then the code uses `numpy`'s `histogram` function (which is equivalent to the current `dev` branch). If `histogram_deposition` is set to `cic` (which is now the default), the code uses a custom-written cython function. I had to use a cython function because `numpy` does not have an available method for CIC deposition.

Here are is the comparison of 1D NGP and CIC histogramming for evenly-spaced particles (the position of the particles is represented by red dots.)
histogram_1d

Here is a 2D comparison
histogram_2d

Many thanks to @jlvay for suggesting to implement this feature!

### DIFF
--- a/opmd_viewer/openpmd_timeseries/plotter.py
+++ b/opmd_viewer/openpmd_timeseries/plotter.py
@@ -8,6 +8,7 @@ Copyright 2015-2016, openPMD-viewer contributors
 Author: Remi Lehe
 License: 3-Clause-BSD-LBNL
 """
+import numpy as np
 try:
     import warnings
     import matplotlib
@@ -76,7 +77,7 @@ class Plotter(object):
            Extent of the histogram
 
         **kw : dict, otional
-           Additional options to be passed to matplotlib's hist
+           Additional options to be passed to matplotlib's bar function
         """
         # Check if matplotlib is available
         check_matplotlib()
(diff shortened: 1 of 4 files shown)


opmd_viewer/openpmd_timeseries/plotter.py
```diff
@@ -8,13 +8,19 @@
...
+import numpy as np
 try:
     import warnings
     import matplotlib
     import matplotlib.pyplot as plt
     matplotlib_installed = True
 except ImportError:
     matplotlib_installed = False
+try:
+    from .cython_function import histogram_cic_1d, histogram_cic_2d
```
ax3l: just FYI on the risks of `try import` of `Cython` without using the import during the try block: https://github.com/pypa/setuptools/issues/1229

RemiLehe:
Thanks for the advice. I updated the `setup.py` accordingly.

PR #187, opened 2018-02-06
Close #186: Adapt the code for openPMD 1.1.0

### LINKED ISSUE #186: Implement optional particle and mesh path
See https://github.com/openPMD/openPMD-standard/pull/171#pullrequestreview-90887938

RemiLehe:
In openPMD 1.1.0, the `meshesPath` and `particlesPath` are optional (indicating that there are no fields/particles, if these attributes are not set).

This adapts the viewer, so that it does not fail when this is the case. 

Note that, with this pull request, the current version of the `openPMD-viewer` can read both openPMD 1.0 and openPMD 1.1.


RemiLehe:
Note: I tested this with files writtten by Warp, using an updated version that writes openPMD 1.1.

PR #188, opened 2018-02-06
Adapt the code for openPMD 1.1.0

RemiLehe:
In openPMD 1.1.0, the meshesPath and particlesPath are optional (indicating that there are no fields/particles, if these attributes are not set).

This adapts the viewer, so that it does not fail when this is the case.

Note that, with this pull request, the current version of the openPMD-viewer can read both openPMD 1.0 and openPMD 1.1.

### DIFF
--- a/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
+++ b/opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
@@ -67,8 +67,12 @@ def read_openPMD_params(filename, extract_parameters=True):
             params['extensions'].append(extension)
 
     # Find out whether fields are present and extract their metadata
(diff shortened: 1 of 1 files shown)


opmd_viewer/openpmd_timeseries/data_reader/params_reader.py
```diff
@@ -67,8 +67,13 @@ def read_openPMD_params(filename, extract_parameters=True):
             params['extensions'].append(extension)
 
     # Find out whether fields are present and extract their metadata
-    meshes_path = f.attrs['meshesPath'].decode().strip('/')
-    if meshes_path in bpath.keys():
+    # Find out whether fields are present
```
ax3l: doubled comment line?

PR #190, opened 2018-05-01
More options to get_emittance

MaxThevenet:
This small PR proposes a more complete version of `get_emittance`. The user can
- choose normalized or trace-space emittance
- choose projected or slice emittance
- get emittance along propagation within one function call.

Function `get_emittance` was modified.
Function `emittance_from_coord` was added.

### DIFF
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -247,23 +247,38 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
         return( div_x, div_y )
 
     def emittance_from_coord(self, x, y, ux, uy, w):
+        """ Calculate emittance from arrays of particle coordinates.
+
+        Parameters
+        ----------
+        x : arrays of floats
+            x position of particles
+        y : arrays of floats
+            y position of particles
+        ux : arrays of floats
+            ux normalized momentum of particles
+        uy : arrays of floats
+            uy normalized momentum of particles
+        w : arrays of floats
+            Particle weights
+        """
         xm = w_ave( x, weights=w )
-        xsq = w_ave( (x-xm) ** 2, weights=w )
+        xsq = w_ave( ( x - xm ) ** 2, weights=w )
         ym = w_ave( y, weights=w )
-        ysq = w_ave( (y-ym) ** 2, weights=w )
-        uxm    = w_ave( ux, weights=w )
-        uxsq   = w_ave( (ux-uxm) ** 2, weights=w )
-        uym    = w_ave( uy, weights=w )
-        uysq   = w_ave( (uy-uym) ** 2, weights=w )
-        xux    = w_ave( (x-xm) * (ux-uxm), weights=w )
-        yuy    = w_ave( (y-ym) * (uy-uym), weights=w )
+        ysq = w_ave( ( y - ym ) ** 2, weights=w )
+        uxm = w_ave( ux, weights=w )
+        uxsq = w_ave( ( ux - uxm ) ** 2, weights=w )
+        uym = w_ave( uy, weights=w )
+        uysq = w_ave( ( uy - uym ) ** 2, weights=w )
+        xux = w_ave( ( x - xm ) * (ux - uxm ), weights=w )
+        yuy = w_ave( ( y - ym ) * ( uy - uym ), weights=w )
         emit_x = ( abs(xsq * uxsq - xux ** 2) )**.5
         emit_y = ( abs(ysq * uysq - yuy ** 2) )**.5
         return emit_x, emit_y
 
-    def get_emittance(self, t=None, iteration=None, species=None, 
-                      select=None, kind='normalized', nslices=0, 
-                      beam_length=None, sum_slices=True, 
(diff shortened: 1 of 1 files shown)


RemiLehe:
Thanks a lot for this PR!
I'll try to have a close look at it in the next few days.
For now, though, I had a quick question: what is the typical use case for using `sum_slices=True` and `nslices>1`? Is that (more or less) equivalent to not using `nslices` at all?

soerenjalas:
Hi,
Sorry for highjacking this discussion a little bit…
First, thanks for the PR! I really like the option to differentiate between trace space and norm. emittance. And the slice emittance stuff is very nice!

Some comments:
* I think having the option to analyse the full time series in a single function call is very useful. However, I think implementing it in every single function makes the code unnecessary bulky. For me having a method like
```python
OPMDobject.analyse_timeseries(property_list=['a0', 'gamma', 'emittance', 'arbitrary self defined function'])
```
would be nice for that purpose. A general problem might be that in many cases the selection filters have to be adjusted along the series, e.g. because the bunch energy increases.

* I assume the slicing is done in energy. Could you make this clear in the documentation string.

* I think the argument `sum_slices` is misleading, maybe `average_slices` is more clear (If I understood the argument correctly)

* The slices width is defined by nslices, but since the energy domain may change over the time series this width would also change. Would it make more sense to define the energy interval?

* Is there a specific reason for having the function `emittance_from_coord`?  If it is just a helper function for `get_emittance`, I would rather define it outside of the class just like `w_std`and `w_avg`.

All these points a clearly just my opinion and are open for debate. 

Thanks

MaxThevenet:
I'll update the PR to include suggestions you both gave and we'll discuss it then.

RemiLehe:
Ok, thanks!
Btw, we could discuss the `all_iterations` issue here: https://github.com/openPMD/openPMD-viewer/issues/191

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -258,45 +259,111 @@ def get_emittance( self, t=None, iteration=None, species=None,
...
+            'z' : [0, 100] (Particles having x between 0 and 100 microns).
+
+        kind : string, optional
+            Kind of emittance to be computed. Cam be 'normalized' or 'trace'.
+
+        type : string, optional
+            Type of emittance to be computed. Can be 'projected' for
+            projected emittance, 'all-slices' for emittance within slices
+            taken along the z direction or 'slice-averaged' for slice
+            emittance averaged over all slices.
```
RemiLehe: Could you format this like that (just for clarity):
```
            Type of emittance to be computed. Available options:
               - 'projected' for projected emittance
               - 'all-slices' for emittance within slices taken along the z direction
               - 'slice-averaged' for slice emittance averaged over all slices.
```

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -258,45 +259,111 @@ def get_emittance( self, t=None, iteration=None, species=None,
...
+
+        nslices : integer, optional
+            Number of slices to compute slice emittance. Required if
+            type='slice-average' or 'all-slices'.
+
+        beam_length : fload (in meters), optional
```
RemiLehe: I think that `fload` should be replaced by `float`

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -258,45 +259,111 @@ def get_emittance( self, t=None, iteration=None, species=None,
...
+        if kind == 'trace':
+            ux = ux / uz
+            uy = uy / uz
+
+        if do_slice_emittance:
+            # Get slice locations
+            zavg = np.average(z, weights=w)
+            z = z - zavg
+            if beam_length is None:
+                std = np.std(z)
```
RemiLehe: Maybe `np.std` should be replaced by a function that uses the particle weigths

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -240,16 +240,17 @@ def get_divergence( self, t=None, iteration=None, species=None,
...
+        # Calculate divergence
         div_x = w_std( np.arctan2(ux, uz), w )
         div_y = w_std( np.arctan2(uy, uz), w )
         # Return the result
         return( div_x, div_y )
 
-    def get_emittance( self, t=None, iteration=None, species=None,
-                       select=None ):
+    def get_emittance(self, t=None, iteration=None, species=None,
+                      select=None, kind='normalized', type='projected',
```
RemiLehe: Actually, I know realize that using `type` is a bad idea (since it's also a Python command). Sorry for suggesting this in the first place.

Could you maybe use `slicing` instead? (and replace throughout the function)

RemiLehe:
@soerenjalas Do you agree with the current version of the PR? If yes, I will merge it.

MaxThevenet:
Hi,
I've made changes to comply with your comments:
- `slice_weight` is the number of electrons in each slice. Also changed docstring
- went back to `w_ave` instead of `w_std`
- use `current_slice` instead of `slice`

Issue #193, opened 2018-05-04
Allow field lineout in addition to slicing?


MaxThevenet:
At the moment, users can get a slice of a field with `get_field(slicing=0., slicing_dir='y')`. Is there a way to get field lineouts? Do you think it could also be an option in OpenPMD? 

I was considering something like `get_field(lineout={'x':0., 'y':0.}, lineout_dir='z')`, or `get_field(lineout={'x':0., 'y':0.})` to avoid redundancy.
I see two ways to do it:
- change functions `read_field_3d`, `read_field_2d` and `read_field_circ` so that they take a new argument
- change `get_field` only so that it takes a slice and then extracts a lineout.

Do you have an opinion?

RemiLehe:
I agree that it would be great to have the possibility to easily to lineouts.

If possible it would be great if the lineout API is compatible with the `slicing` API. For instance, what is your opinion about doing using:
```
ts.get_field( slicing=[0., 0.], slicing_dir=['x', 'y'] )
```
in order to get a lineout from 3D data?

MaxThevenet:
Sure, that's great.

PR #195, opened 2018-05-05
Allow slicing along several directions

MaxThevenet:
This PR aims at allowing slicing along several directions in 3d and 2d. It answers the [issue on field lineouts](https://github.com/openPMD/openPMD-viewer/issues/193).
Users can pass lists instead of single elements for slicing arguments:
`F, infoF = ts.get_field( ... slicing = [0.0, 0.2], slicing_dir=['z', 'x'])`
This change is intended to be backward-compatible, and 
`F, infoF = ts.get_field( ... slicing = 0.0, slicing_dir='z')`
gives the same result as
`F, infoF = ts.get_field( ... slicing = [0.0], slicing_dir=['z'])`
Main changes were done in function `get_data`.

Note that 2d and circ datasets can be sliced along 0, 1 or 2 direction, which returns respectively a 2d array, a 1d array and a single element.

Also, `read_field_1d`, `read_field_2d` and `read_field_3d` have been merged to read_field_cartesian, since they all allow slicing and `get_data` is general enough.


MaxThevenet:
This is [WIP] as it requires more testing on infoF. Also current version breaks the slider.

MaxThevenet:
This PR should be ready for review!

opmd_viewer/openpmd_timeseries/main.py
```diff
@@ -347,7 +347,7 @@ def get_particle(self, var_list=None, species=None, t=None, iteration=None,
             return(data_list)
 
     def get_field(self, field=None, coord=None, t=None, iteration=None,
-                  m='all', theta=0., slicing=0., slicing_dir='y',
+                  m='all', theta=0., slicing=None, slicing_dir=None,
```
MaxThevenet: Having `y` as the default slicing direction was kind of arbitrary. Now that slicing is also allowed in 2d, it can lead to an issue.
Depending on the code, 2d simulations can rely on any set of 2 coordinates in `x,y,z`. If the code uses `x-z` as coordinates for 2d simulations, `get_field` returns a 2d array by default. If the code uses `x-y` instead, `get_field` returns a 1d array by default.
To avoid this issue, @RemiLehe and I agreed to change the behavior of `get_field` so that no slicing is done by default. However, it breaks backward compatibility, so it may impact some users.

@soerenjalas @hightower8083 @ax3l is this fine for you?

RemiLehe: Just to clarify @MaxThevenet's comment: the backward-compatibility breaks exclusively in the case where the user calls `get_field` **without explicitly setting the `slicing_dir` argument** and **for a 3D dataset**. 

For instance
```
E, info = ts.get_field('E', 'x')
```
used to return `E` as a 2d array (arbitrarily sliced along 'y'), but with this PR it returns a 3d array (and the user needs to explicitly set `slicing_dir` in order to get a 2d slice).

For 2d and circ dataset, the backward-compatibility is preserved.

hightower8083: No problem with me -- the proposed behaviour is intuitively correct. 
By the way, for my curiosity, are there plans for the z-silcing in the multi-mode circ? 
Its probably kinda a rarely needed thing (guess can be done in 3d via Visit), but nevertheless at some point this could be nice to have, in particular to see propagation of laser with a realistic profile.

ax3l: That's fine, just make sure to release it into a `0.9` or `1.0` version so people see the possible break.

@MaxThevenet you can also already add this breaking behavior to the `CHANGELOG.md` file.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -410,10 +410,25 @@ def get_laser_envelope( self, t=None, iteration=None, pol=None, m='all',
...
            The angle of the plane of observation, with respect to the x axis
 
-        slicing_dir : str, optional
-           Only used for 3dcartesian geometry
-           The direction along which to slice the data
-           Either 'x', 'y'
+        slicing : float or list of float, optional
+           Number(s) between -1 and 1 that indicate where to slice the data,
+           along the directions in `slicing_dir`
+           -1 : lower edge of the simulation box
```
ax3l: Just a comment: that's an interesting choice instead of a `[0.0 : 1.0]` range for the relative simulation box. Is there a specific reason for that?

MaxThevenet: I kept it the same as before the PR. Both are fine for me.

ax3l: Ah I see, alright! Yes I personally prefer a length-one range with default 0.5 usually.
@RemiLehe since we break the API anyway, do we want to "simplify" the relative simulation box range on the way?

RemiLehe: Yeah, actually I think that I initially choose `-1` to `1` because the direction for slicing is oftentimes a transverse direction, where the bounds are symmetrical (e.g. `x` from -50 microns to +50 microns). In that case, I thought that it would be more intuitive for a user to have these bounds correspond to `-1` and `+1`. 

But it's true that this assumes a particular kind of simulation... We could change, but note that this extends the non-backward-compatibility, in the sense that 
```
get_field( 'E', 'x', slicing_dir='y', slicing=0)
```
is backward-compatible with the current PR, but would break with the proposed change.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -410,10 +410,25 @@ def get_laser_envelope( self, t=None, iteration=None, pol=None, m='all',
...
+           0 : middle of the simulation box
+           1 : upper edge of the simulation box
+           If slicing is None, the full grid is returned.
```
RemiLehe: I think that 
```
If slicing is None, the full grid is returned.
Default is None.
```
Should be replaced by 
```
If slicing_dir is None ...
```
and should be moved to the description of `slicing_dir`. Is that correct?

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -629,16 +644,23 @@ def get_spectrum( self, t=None, iteration=None, pol=None,
         if pol not in ['x', 'y']:
             raise ValueError('The `pol` argument is missing or erroneous.')
         if pol == 'x':
-            slicing_dir = 'y'
             theta = 0
         else:
-            slicing_dir = 'x'
             theta = np.pi / 2.
+        if "3dcartesian" in self.avail_geom:
+            slicing = 0.
```
RemiLehe: The above line (`slicing = 0.`) should probably be removed.

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -689,16 +711,24 @@ def get_a0( self, t=None, iteration=None, pol=None ):
...
+            if pol == 'x':
+                slicing_dir = 'y'
+            else:
+                slicing_dir = 'x'
+        else:
+            slicing = None
```
RemiLehe: Why is it important to set `slicing` in the above code?  Shouldn't the default `slicing` give pretty much the same result?

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -738,14 +768,22 @@ def get_ctau( self, t=None, iteration=None, pol=None, method='fit' ):
...
```
RemiLehe: Same question on the default `slicing`

opmd_viewer/openpmd_timeseries/data_reader/field_reader.py
```diff
@@ -173,90 +211,52 @@ def read_field_circ( filename, field_path, m=0, theta=0. ):
...
-            group.attrs['gridSpacing'], group.attrs['gridGlobalOffset'],
-            group.attrs['gridUnitSI'], dset.attrs['position'] )
+
+    # Perform slicing if needed
+    if slicing_dir is not None:
+        # Convert arguments to lists
+        if not isinstance(slicing, list):
+            slicing = [slicing]
+        if not isinstance(slicing_dir, list):
+            slicing_dir = [slicing_dir]
```
RemiLehe: The above could be simply:
```
if slicing_dir is not None:
    slicing = list(slicing)
    slicing_dir = list(slicing_dir)
```

opmd_viewer/openpmd_timeseries/data_reader/field_reader.py
```diff
@@ -173,90 +211,52 @@ def read_field_circ( filename, field_path, m=0, theta=0. ):
...
+            ind = slicing_dir.index('r')
+            n_cells = shape[ 0 ]
+            i_cell = int( 0.5 * (slicing[ind] + 1.) * n_cells )
+            i_cell = max( i_cell, 0 )
+            i_cell = min( i_cell, n_cells - 1)
+            F_total = F_total[i_cell, ...]
+            shape.pop( ind )
+            grid_spacing.pop( ind )
+            global_offset.pop( ind )
+            axis_labels = axis_labels[1:]
```
RemiLehe: Could you generalize the above code (up to `if 'z' in slicing_dir:`) with a loop over `slicing_dir` (just like in the cartesian case)? 

For the slicing, I think you could do 
```
F_total = np.take( F_total, i_cell, axis=ind )
```

opmd_viewer/addons/pic/lpa_diagnostics.py
MaxThevenet: Right, this no longer make sense with these default values.

opmd_viewer/openpmd_timeseries/data_reader/field_reader.py
```diff
@@ -26,81 +27,106 @@ def read_field_1d( filename, field_path, axis_labels ):
...
+        slicing = [slicing]
+    if slicing_dir is not None and not isinstance(slicing_dir, list):
+        slicing_dir = [slicing_dir]
+    # Pop slicing directions in arguments that are not in axis_label
+    if slicing_dir is not None:
+        ind_to_pop = [slicing_dir.index(val) for val in slicing_dir if
+            val not in axis_labels ]
+        for ind in ind_to_pop:
+            slicing_dir.pop(ind)
+            slicing.pop(ind)
```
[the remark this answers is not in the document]
MaxThevenet: agreed

opmd_viewer/openpmd_timeseries/data_reader/utilities.py
```diff
@@ -73,24 +73,41 @@ def get_data(dset, i_slice=None, pos_slice=None, output_type=np.float64):
     --------
     An np.ndarray (non-constant dataset) or a single double (constant dataset)
     """
+    # For back-compatibility: Convert pos_slice and i_slice to
+    # single-element lists if they are not lists (e.g. float
+    # and int respectively).
+    if pos_slice is not None and not isinstance(pos_slice, list):
+        pos_slice = [pos_slice]
+    if i_slice is not None and not isinstance(i_slice, list):
+        i_slice = [i_slice]
```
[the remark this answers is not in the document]
MaxThevenet: This works with string but not with floats

MaxThevenet:
Thanks @RemiLehe for your comments, they are taken into account in the last commit.

ax3l:
@MaxThevenet looks like there is a little conflict in the tutorial files you changed now due to our recent update in #202 . Can you please rebase?

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -410,10 +410,25 @@ def get_laser_envelope( self, t=None, iteration=None, pol=None, m='all',
...
+           Default is None
+
+        slicing_dir : str or list of str, optional
+           Direction(s) along which to slice the data
+           + In cartesian geometry, elements can be:
+               - 1d: 'z'
+               - 2d: 'x' and/or 'z'
+               - 3d: 'x' and/or 'y' and/or 'z'
+               - 1d/circ: not implemented
+           + In cylindrical geometry, elements can be 'r' and/or 'z'
```
RemiLehe: The above line seems to contradict `1d/circ: not implemented`. Did I misunderstand?

opmd_viewer/openpmd_timeseries/data_reader/field_reader.py
```diff
@@ -173,90 +205,79 @@ def read_field_circ( filename, field_path, m=0, theta=0. ):
...
+#             ind = slicing_dir.index('r')
+#             n_cells = shape[ 0 ]
+#             i_cell = int( 0.5 * (slicing[ind] + 1.) * n_cells )
+#             i_cell = max( i_cell, 0 )
+#             i_cell = min( i_cell, n_cells - 1)
+#             F_total = F_total[i_cell, ...]
+#             shape.pop( ind )
+#             grid_spacing.pop( ind )
+#             global_offset.pop( ind )
+#             axis_labels = axis_labels[1:]
```
RemiLehe: Could you remove the above commented lines?

MaxThevenet:
I fixed the slicing in circ mode, it should be ready to merge.

RemiLehe:
Thanks for the changes!
I added a few corrections/modifications, and improved the slider support. 
The slider still breaks for cylindrical data, but I'll try to fix this today.

PR #205, opened 2018-05-18
Quick fix for Windows installation

AngelFP:
Installation using pip fails on Windows if the path in line 51 incudes a final slash './' .
Removing it fixes this issue and installation on Linux shouldn't be affected (works fine in Ubuntu).


RemiLehe:
Great! Thanks a lot for fixing this! 
At some point, I should set up continuous integration for Windows...

Issue #206, opened 2018-05-18
Add support and continuous integration on Windows


RemiLehe:
We could use AppVeyor for CI on Windows...

PR #207, opened 2018-05-23
Fix Windows issues

AngelFP:
Small fixes for Windows-specific bugs, all of them related to how Windows handles file paths.

The use of `os.path.join` for navigating within data files has been substituted by manually adding slashes. The issue is that this function uses a double backlash `\\` as separator in Windows, which is appropriate for file paths but not for navigating within data files, where a `/` is still required.


ax3l:
Thank you for the PR!
Can you please remove the `os` import in `main.py` and `data_reader/` files to pass the `flake8` check? :)

Related to #206

opmd_viewer/openpmd_timeseries/data_reader/field_reader.py
```diff
@@ -288,11 +288,11 @@ def find_dataset( dfile, field_path ):
     relative_meshes_path = dfile.attrs["meshesPath"].decode()
 
     # Get the proper dataset
-    full_field_path = os.path.join(base_path, relative_meshes_path, field_path)
+    full_field_path = base_path + '/' + relative_meshes_path + '/' + field_path
```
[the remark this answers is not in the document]
ax3l: (alternatively, we can write our own, safe `join()` method that even allows malformed paths)

opmd_viewer/openpmd_timeseries/data_reader/utilities.py
```diff
@@ -124,3 +124,34 @@ def get_shape(dset):
         shape = dset.shape
 
     return(shape)
+
+
+def join(base_path, *paths):
```
AngelFP: In the end I implemented a simple join method which I think should be enough for this case.

RemiLehe: I think that's a great way to do it!
One very minor comment: could you name this function `filepath_join` instead of `join`, so that the name tells a bit more about what the function does.

opmd_viewer/openpmd_timeseries/data_reader/utilities.py
```diff
@@ -124,3 +124,34 @@ def get_shape(dset):
...
+    A string with the complete path using '/' as separator.
+    """
+    # Initialize path
+    path = base_path
+    # Join all path components adding a '/'
+    try:
+        for p in paths:
+            path += '/' + p
+    except:
+        print("Error joining strings.")
```
RemiLehe: I think that the above code could be made a bit more concise:
```python
def filepath_join( *paths ):
     """
     Copy previous docstring... but without base_path
     """
     path = '/'.join(paths)
     path = path.replace('//', '/')
     return path
```
Would you be okay with using the above version in your code?

opmd_viewer/openpmd_timeseries/main.py
```diff
@@ -460,7 +459,7 @@ def get_field(self, field=None, coord=None, t=None, iteration=None,
             field_path = field
             field_label = field
         elif self.fields_metadata[field]['type'] == 'vector':
-            field_path = os.path.join(field, coord)
+            field_path = field + '/' + coord
```
RemiLehe: Could you also use the new function `filepath_join` here?

RemiLehe:
Thanks for this nice PR! I have a few minor comments (see inline code)

opmd_viewer/openpmd_timeseries/data_reader/utilities.py
RemiLehe: Sounds good. Or maybe even `get_infile_path`.

ax3l: `join_infile_path`?

AngelFP: I like this last name. I just submitted a new commit including both changes.

opmd_viewer/openpmd_timeseries/main.py
AngelFP: done!

PR #211, opened 2018-05-26
Remove microns and femtoseconds

RemiLehe:
A long-requested feature of `openPMD-viewer` is to have consistent units throughout all the function calls (e.g. meters and seconds everywhere, instead of the occasional microns and femtoseconds).

This change would go into the upcoming openPMD-viewer 1.0, which has a breaking API change.

Note that this conversion makes plotting less readable (because matplotlib typically prints 1.e-6 as 0.000001). However, I will address this issue in a later pull request.

### DIFF
--- a/opmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/opmd_viewer/addons/pic/lpa_diagnostics.py
@@ -69,8 +69,8 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
         select : dict, optional
             Either None or a dictionary of rules
             to select the particles, of the form
-            'x' : [-4., 10.]   (Particles having x between -4 and 10 microns)
-            'z' : [0, 100] (Particles having x between 0 and 100 microns)
+            'x' : [-4., 10.]   (Particles having x between -4 and 10)
+            'z' : [0, 100] (Particles having x between 0 and 100)
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -160,12 +160,12 @@ def get_sigma_gamma_slice(self, dz, t=None, iteration=None, species=None,
         if plot:
             check_matplotlib()
             iteration = self.iterations[ self._current_i ]
-            time_fs = 1.e15 * self.t[ self._current_i ]
+            time = self.t[ self._current_i ]
```
ax3l: maybe it's still useful to see the time is already in SI here, e.g. varname `time_s` ?

opmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -69,8 +69,8 @@ def get_mean_gamma( self, t=None, iteration=None, species=None,
         select : dict, optional
             Either None or a dictionary of rules
             to select the particles, of the form
-            'x' : [-4., 10.]   (Particles having x between -4 and 10 microns)
-            'z' : [0, 100] (Particles having x between 0 and 100 microns)
+            'x' : [-4., 10.]   (Particles having x between -4 and 10)
```
ax3l: Maybe keep hints for the users explicit?
here: meters

PR #215, opened 2018-06-07
Better formatting for axes ticks

RemiLehe:
The default matplotlib formatter for axes has one shortcoming: numbers on the order of 1e-6 are printed as 0.000001. With the recent PR #211 , this becomes a problem because, for e.g. laser plasma acceleration, most distances are of this order of magnitude.

Consider for instance this example:
```
vals = 1. + 3.e-6*np.random.rand(400)
plt.hist( vals )
```
Which produces this result, where labels are quite hard to read.
screen shot 2018-06-07 at 2 30 38 pm

Note however that the default formatter has a nice property: any constant offset is detected and printed on the side of the axis (`+1` in this case.)

It would be nice to retain the above property, while having a better way to print numbers on the order of 1e-6.

This PR addresses the issue by creating a subclass of the default matplotlib formatter, which retains the offset, but prints the numbers in scientific format, with exponents that are multiples of 3. This helps in quickly identifying the closest units (e.g. microns, nanometers, etc.)

In the above example, one gets
screen shot 2018-06-07 at 2 34 45 pm

### DIFF
--- a/opmd_viewer/openpmd_timeseries/plotter.py
+++ b/opmd_viewer/openpmd_timeseries/plotter.py
@@ -9,6 +9,7 @@ Author: Remi Lehe
 License: 3-Clause-BSD-LBNL
 """
 import numpy as np
+import math
 try:
     import warnings
     import matplotlib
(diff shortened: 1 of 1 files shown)


RemiLehe:
This is way of printing numbers is clearly a choice. Any alternative suggestion is also welcome. 

Note also that, at some point (for very small numbers), matplotlib will show the order of magnitude on the side:
screen shot 2018-06-07 at 2 37 55 pm

However, I find this somewhat less readable. But maybe it is a matter of taste.

PR #217, opened 2018-06-20
Improve printed time in plot title

RemiLehe:
Now that (on the branch `upcoming-1.0`) the time in seconds everywhere, the title on the plots shows `t = 0.0 s` all the time, because the number is too small (~1.e-15) to be printed with `%f`. Printing with `%e` solves the issue.

### DIFF
--- a/opmd_viewer/openpmd_timeseries/plotter.py
+++ b/opmd_viewer/openpmd_timeseries/plotter.py
@@ -171,7 +171,7 @@ class Plotter(object):
         plt.xlim( hist_range[0] )
         plt.ylim( hist_range[1] )
         plt.xlabel(quantity1, fontsize=self.fontsize)
-        plt.title("%s:   t =  %.0f s    (iteration %d)"
+        plt.title("%s:   t =  %.2e s    (iteration %d)"
                   % (species, time, iteration), fontsize=self.fontsize)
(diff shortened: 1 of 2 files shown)


opmd_viewer/openpmd_timeseries/plotter.py
```diff
@@ -354,20 +354,14 @@ def show_field_2d(self, F, info, slicing_dir, m, field_label, geometry,
...
+        else:
+            plt.title("%s at %.2e s   (iteration %d)"
                       % (field_label, time, iteration),
                       fontsize=self.fontsize)
-        # 3D Cartesian geometry
-        elif geometry == "3dcartesian":
-            slice_plane = info.axes[0] + '-' + info.axes[1]
-            plt.title("%s sliced in %s at %.1f s  (iteration %d)"
-                      % (field_label, slice_plane, time, iteration),
-                      fontsize=self.fontsize)
```
RemiLehe: I also suppressed the slicing information that was specific for the 3D case, since 2D and circ data can now also be sliced.

ax3l:
I always wonder if exponential formatting in bases of 10^3 exists. This would be so useful for physical quantities (fs, ps, ...). But besides that :+1:

PR #218, opened 2018-09-13
Allow slice emittance calculation along arbitrary coordinates

soerenjalas:
Often it can be useful to calculate the slice emittance for along another coordinate than z. I changed the method so that now all particle properties can be used for slicing e.g. to calculate the energy (actually uz) dependent emittance. 

There are a few API breaking changes that we would have to decide on (they can easily be reversed)

* the input argument beam_length has now the more general name slice_range
* the slice centers are returned as a FieldMetaInformation object

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,8 @@ before_install:
 
 # Install packages
 install:
-  - conda install --yes numpy >=1.10
-  - conda install --yes numpy cython numpy scipy h5py
+  - pip install numpy
+  - conda install --yes cython  scipy h5py
   - pip install matplotlib
   - python setup.py install
   - pip install wget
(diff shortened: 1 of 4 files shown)


ax3l:
@soerenjalas can you please rebase, remove already merged commits and squash together commits that change the same aspect?

soerenjalas:
Ahhh... this is getting out of hand and I'm too much of a git noob. I will create a new branch and PR, sorry.

RemiLehe:
OK, thanks @soerenjalas. Could actually wait until I update `upcoming-1.0` and then start your new branch from that?
Thanks!

RemiLehe:
@soerenjalas I just updated the `upcoming-1.0` branch. You can use it now.

PR #222, opened 2018-09-29
Fix slider when using Jupyterlab

RemiLehe:
This pull request fixes the slider when using JupyterLab.

The issue was that JupyterLab does not support the `%matplotlib notebook` magic. Instead JupyterLab uses the `%matplotlib widget` magic (which requires the [jupyter-matplotlib extension](https://github.com/matplotlib/jupyter-matplotlib)).

However, the rules that determines whether a figure will show up in the notebook are different with `%matplotlib widget` (namely, the figure will only show up if this figure is created for the first time. In other words, running the cell `plt.figure(4)` will show figure number 4 only the first time that this cell is run ; later execution will not show the figure).
These different rules meant that the slider was not showing the figures when using `%matplotlib widget`. This was fixed by including an explicit `display` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,9 @@ distribution](https://www.continuum.io/downloads), and then type
 ```
 conda install -c rlehe openpmd_viewer
 ```
+If you are using JupyterLab, please also install the `jupyter-matplotlib`
+extension (See installation instructions
+[here](https://github.com/matplotlib/jupyter-matplotlib)).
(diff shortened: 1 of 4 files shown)


RemiLehe:
@ax3l Yes, I think that this is a good idea! I would do a separate PR for this, though, since it would need some testing of the installation, the conda build, etc.

PR #231, opened 2019-05-19
Remove pep8 checks

RemiLehe:
Although checking PEP8 is useful in some cases, my impression is that, in the case of openPMD-viewer, this slows down code development because of minor styling conventions (e.g. spaces between operators, etc.)

I would be in favor of removing the PEP8 checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,14 +36,12 @@ install:
   - pip install matplotlib
   - python setup.py install
   - pip install wget
-  - pip install pyflakes pycodestyle jupyter pytest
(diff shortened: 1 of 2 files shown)


ax3l:
Maybe we all should just use [black](https://github.com/python/black) ;-)

ax3l:
One could also automate the style updates, e.g.: https://github.com/OrkoHunter/pep8speaks

RemiLehe:
Good point @ax3l. We could look into this in the future.

PR #232, opened 2019-05-19
Reconstruct full 3D fields from cylindrical simulation

RemiLehe:
This allows `openPMD-viewer` to return a 3D cartesian array, from a file contains the `thetaMode` representation of the fields.
More specifically, `ts.get_field` will return a numpy 3darray (and corresponding `info` object) if the user passes `theta=None` (note that the default value for `theta` is 0)

This feature can be useful for postprocessing operations that are easier on a 3D cartesian grid (e.g. computing the total energy). One of the motivations of this PR is also to allow [VisualPIC](https://github.com/AngelFP/VisualPIC) to [work with cylindrical data](https://github.com/AngelFP/VisualPIC/issues/11) (including 3D visualization of the fields).

Regarding the implementation:
- I updated the functions that read the field in quasi-3D, to support `theta=None`.
- I added an example in the tutorial notebook.
- Note that right now, the function `construct_3d_from_circ` typically dominates the time to get the field. I tried to speed up this function using Cython (see [this commit](https://github.com/openPMD/openPMD-viewer/pull/232/commits/7bdd3cec36db0e286a0f7f49fbfb65b3bd0f7624)), but this made it 10x **slower**. I obviously did something wrong but I am not sure what. So in the end, this function remained in Python/Numpy in this PR.


opmd_viewer/openpmd_timeseries/data_reader/field_reader.py
```diff
@@ -138,40 +142,64 @@ def read_field_circ( filename, field_path, m=0, theta=0. ):
...
     else:
-        # Extract higher mode
-        cos = np.cos( m * theta )
-        sin = np.sin( m * theta )
-        F_cos = get_data( dset, 2 * m - 1, 0 )
-        F_sin = get_data( dset, 2 * m, 0 )
-        F = cos * F_cos + sin * F_sin
-        F_total[Nr:, :] = F[:, :]
-        F_total[:Nr, :] = (-1) ** m * F[::-1, :]
+
```
RemiLehe: The code below is unchanged (just reindented)

RemiLehe:
Thanks for the feedback.
In fact this is doing nearest-grid-point interpolation (so not really an interpolation, as you noted). But I'll maybe replace it by linear interpolation in a separate PR.

PR #240, opened 2019-07-27
Replace Module name from `opmd_viewer` to `openpmd_viewer`

RemiLehe:
As discussed with @ax3l and @soerenjalas, in version 1.0, we will replace
```
import opmd_viewer
```
by 
```
import openpmd_viewer
```
This is what the current PR does, by:
- `git mv`-ing the whole `opmd_viewer` folder to `openpmd_viewer`
- `sed`-ing all relevant files to automatically change `opmd_viewer` into `openpmd_viewer`

There will also be a later PR that leaves a stub, so that importing `opmd_viewer` will raise an Exception (prompting the user to either downgrade to version `0.X` or to change the import syntax.)


ax3l:
Thx! I will have a detailed look on it later today.

I renamed the PR accordingly, since the "python import"-name is the "module" name. The package name would be the PyPI and conda-forge name.

Ref.: https://github.com/openPMD/openPMD-api/issues/181#issuecomment-427859010

PR #241, opened 2019-07-31
Drop python 3.4 support in CI

RemiLehe:
It seems that python 3.4 is no longer supported by conda:
https://travis-ci.org/openPMD/openPMD-viewer/jobs/566097165


ax3l:
Ok, but can we please add Python 3.7 tests at the same time? :)
Python 3.8 is around the corner :)

PR #243, opened 2019-08-23
[WIP] Add deprecation warning when using `opmd_viewer`

RemiLehe:


opmd_viewer/__init__.py
```diff
@@ -0,0 +1,29 @@
+"""
+This is a stub that detects whether the user is attempting to use the
+old import statement from openPMD-viewer 0.X (`import opmd_viewer`).
+
+In that case, an exception is raise, that prompts the user to use the new API.
```
ax3l:
```suggestion
In that case, an exception is raised that prompts the user to use the new API.
```

Issue #233, opened 2019-05-20
Replace Cython with Numba


RemiLehe:
Currently, in `openPMD-viewer`, we are using `Cython` in order to speed up some of the functions that are critical to performance and cannot be efficiently expressed with numpy.

However, my personal experience is that writing Cython function is quite a bit of work. And getting good performance (or even any speed-up at all: see #232) is sometimes difficult. Also for my experience: Numba is often much easier to use "out-of-the-box". (In separate tests, I confirmed for instance that the function I tried to accelerate in #232 gets a large speed up with Numba).

In the past, we avoided using Numba in openPMD-viewer because it was typically difficult to install (esp. without `conda`). However, this changed a while ago, when the Anaconda team made it possible to install numba with `pip`.

Thus, I would be in favor of adding Numba as an optional dependency (same as for Cython currently), and replace Cython annotations with Numba JIT decorators. (We could also adde the `cache=True` option in the decorator, so that these functions would not be recompiled everytime the user starts a new Python process.)

@ax3l @hightower8083 @soerenjalas Any opinion?

ax3l:
Numba is fine with me, although it's not a light dependency.
Pythran is also a worthwhile alternative https://github.com/serge-sans-paille/pythran

Anyway, one would install both through conda...

hightower8083:
Well for the real speed-up (and enabling of openMP) there's always `f2py` in `numpy` =) 
but Numba looks really good to me, and its straightforward to replace cython.

StanczakDominik:
Seems like this can now be closed due to #246 getting merged? :)

RemiLehe:
That's correct @StanczakDominik. Thanks for the reminder :)

PR #246, opened 2019-09-08
Use Numba instead of Cython

RemiLehe:
As mentioned in #233, Numba now seems to be easier to install and use than Cython.
Therefore, this PR removes Cython code and replaces it by Numba.


openpmd_viewer/openpmd_timeseries/utilities.py
```diff
@@ -222,7 +223,85 @@ def combine_cylindrical_components( Fr, Ft, theta, coord, info ):
 
     return F
 
-
```
RemiLehe: The functions below used to be defined in the file `cython_functions.pyx`.

PR #247, opened 2019-10-13
Update laser functions with new slicing syntax

RemiLehe:
This PR makes a number of laser functions more consistent with the rest of the `openPMD-viewer` API, and refactors some of the code. More specifically:

- The function `get_laser_envelope` does not support the argument `index` anymore
(which was effectively used in order to perform slicing). Instead, users should use
the argument `slicing_dir` (which is now more fully supported).

- In addition, `get_laser_envelope` now supports the argument `plot_range`. This is done by using standard plotting tools that we also use for `get_field`, instead of customize `matplotlib` commands that were used previously.

- A number of laser-diagnostic functions were broken for 3D datasets, because of the new slicing convention in `openPMD-viewer 1.0`, or used complicated syntax in order to extract a 1D lineout. With the ability of `openPMD-viewer 1.0` to do a 1D slice immediately in `get_field` (and in `get_laser_envelope` as part of this PR), this can be fixed/refactored in a more straight-forward way. This PR uses this, and explicitly takes a 1D lineout along the `z` axis, inside the functions `get_spectrum`, `get_a0`, `get_ctau`, `get_spectrogram`.

- The function `get_laser_waist` was also broken for 3D datasets, and it had a `slicing_dir` argument that was not really supported. This was fixed by removing the `slicing_dir` argument and (arbitrarily) slicing across `y` for 3D datasets.

- Finally, a new helper function `sanitize_slicing` is used internally, in order to more easily handle the flexibility of the `slicing` arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log / Release Log for openPMD-viewer
 
-## 1.0
+## 1.0.0
 
 This version introduces major changes and breaks backward compatibility.
 
@@ -24,6 +24,10 @@ used to return the central slice along `y` while it now returns the full 3d fiel
 - A new function (`ts.iterate`) was introduced in order to quickly apply a
 given function to all iterations of a time series. See the docstring of
 `ts.iterate` for more information.
+- The function `get_laser_envelope` does not support the argument `index` anymore
+(which was effectively used in order to perform slicing). Instead, users should use
+the argument `slicing_dir`. In addition, `get_laser_envelope` now supports the
+argument `plot_range`.
 
 ## 0.9.0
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -21,9 +21,16 @@ x, y, z = ts.get_particle(['x', 'y', 'z'], iteration=1000)
 compatibility for 3d field:
 ```get_field(field=field, coord=coord, iteration=iteration)```
 used to return the central slice along `y` while it now returns the full 3d field.
```
RemiLehe: I forgot to add the lines below in a previous PR ; sorry to include it here as part of this PR.

soerenjalas:
Thanks for the excessive changes. Everything looks good to me!

Issue #248, opened 2019-11-19
Using get_current() for a plasma density profile which starts from vacuum


abonatto:
Dear Developers,

In order to plot the current of plasma electrons along the beam propagation in a plasma, I am using LpaDiagnostics's get_current(). However, for a species which has not yet "been started" (for example, when the beam starts in vacuum and, after a given distance, there is a plasma density up-ramp), I get the following error:
```
ValueError: zero-size array to reduction operation maximum which has no identity
```
 Would you have any suggestion regarding how to deal with this no-plasma region in this situation? I thought of using try, except. However, I believe it would be interesting to have a better way to deal with it.

Cheers,

Alex

ax3l:
Hi Alex,

thank you for the detailed report!
Soeren provided a fix via #250 which is now in `dev` - does that fix it for you?

abonatto:
Hi Axel,

Sorry for taking so long to answer! I uninstalled/reinstalled via pip, using the following command to install the dev branch (please let me know if this is correct):
```
pip install git+https://github.com/openPMD/openPMD-viewer@dev
```
However, I had the same failure.

Cheers,

Alex

ax3l:
Thanks!
Did you make sure it's definitely the new version you are using? e.g. `pip`/`pip3` mismatches, etc. You can verify usually with
```py
import <modulename> as foo
print(foo.__version__)
print(foo.__path__)
```
and check the sources under the path are what you expected.

abonatto:
Yep. To double check, I pip-uninstalled it again, and tried to load it in python to ensure that there were no other versions installed. Then, i reinstalled it manually, using
```
git clone https://github.com/openPMD/openPMD-viewer
git fetch
git checkout dev
python setup.py install
```
After that, I confirmed the path and version using the commands you suggested me. However, I am still experiencing the following error (even when I load a .h5 file in which the plasma density is already initialised/non-zero):
```
In [51]: ts2.avail_species
Out[51]: [u'beam', u'electrons']

In [52]: beam_current = ts2.get_current(iteration=iter, species='beam')

In [53]: eplasma_current = ts2.get_current(iteration=iter, species='electrons')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-53-d19fabc1a152> in <module>()
----> 1 eplasma_current = ts2.get_current(iteration=iter, species='electrons')

[...]

PR #250, opened 2019-11-27
Make get_current more robust for empty particle selections

soerenjalas:
This is a proposal to solve the issue #248. The problem is that np.histogram is not robust against empty inputs. With this PR get_current returns an array of length `bins` filled with zeros if the particle selection is empty.


ax3l:
cc @abonatto

ax3l:
re-opening to trigger CI

PR #255, opened 2019-12-16
Bugfix #248 get_current() for empty particle selections

soerenjalas:
The issue was that np.min() and np.max() were still called on empty arrays. This PR sets the bounds for the `FieldMetaInformation` object to 0 if `get_particle` returns empty particle arrays.


RemiLehe:
Thanks for this fix!

PR #256, opened 2019-12-18
file close() call missing

hightower8083:
we use `check_all_files=False` for huge track outputs, and i'm not sure how safe it is to leave so many files open. let's fix it.


RemiLehe:
Thanks! Good catch!

PR #283, opened 2020-06-16
Initiate ParticleTracker with `id` array instead of a dictionary

hightower8083:
An option to pass `id` array to ParticleTracker through `select=` argument. Useful when need more complex selection than the one provided by standard `select` dictionary.


RemiLehe:
Thanks a lot for adding this!

PR #293, opened 2020-11-25
Add a new variable `openpmd_viewer.backend`

RemiLehe:
With this new variable, the user can now check the backend that is being used:
```py
import openpmd_viewer
print( openpmd_viewer.backend )
```


ax3l (approved):
Oh, that's it? Cool! :)

PR #295, opened 2020-12-04
add backend choice option

hightower8083:
sketch of the backend choice option
**For demonstration purpose, not to merge**

### DIFF
--- a/openpmd_viewer/__init__.py
+++ b/openpmd_viewer/__init__.py
@@ -7,9 +7,9 @@ See the class OpenPMDTimeSeries to open a set of openPMD files
 """
 # Make the OpenPMDTimeSeries object accessible from outside the package
 from .openpmd_timeseries import OpenPMDTimeSeries, FieldMetaInformation, \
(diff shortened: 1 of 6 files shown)


ax3l:
Could one also implement this as a module property?

hightower8083:
@ax3l well I'm not sure we use that much the module properties in python.. 
Can you explain in a few more words  the expected behaviour ?

RemiLehe:
I actually like the approach used here by @hightower8083 to let each `OpenPMDTimeSeries` object have their own backend (with reasonable defaults, i.e. using `openpmd-api` whenever it is available).
This gives more freedom to the user, and allows workflow where they can use different backends within the same notebook.

I proposed some additional changes [in this PR](https://github.com/hightower8083/openPMD-viewer/pull/3).
@hightower8083 If you agree with these changes, you can merge that PR, and then we could consider merging this into the main repository.

hightower8083:
structure looks good to me

PR #289, opened 2020-10-07
Isolate `h5py`-specific code, in preparation for other `openPMD` reader

RemiLehe:
Generalization of the backend routines in preparation for #285

This PR creates a new `DataReader` object. All accesses to the files are now done through this `DataReader` object. 
Right now, the only implemented backend for the `DataReader` object is `h5py` ; but this PR makes it possible to implement other backends (e.g. `openpmd-api`) for the `DataReader` object.

In addition, all the `hp5py`-specific code is now moved into a new folder `openpmd_viewer/openpmd_timeseries/data_reader/h5py_reader/`.

Issue #296, opened 2020-12-04
compatibility issues of new openpmd-api


hightower8083:
Hi @RemiLehe  and @ax3l 

I've tested the new `openpmd-api` integration (using `openpmd-api@0.12.0a` from *conda*), and I have some comments:
- for WarpX it works well for 2D (not tested 3D) for `bp` and `h5` backends
- for WarpX with RZ it shows `cartesian2d` geometry and does not get any modes, but I guess its  (see [#1543](https://github.com/ECP-WarpX/WarpX/issues/1543) )
- for FBPIC it does not work at all returning 
```[HDF5] Internal error: Failed to open HDF5 attribute during attribute read```
- with FBPIC, if I force the `h5py` backend (see #295 ), it can read but only the vector fields (`E`, `B` and `J`), while scalars (`rho` and `rho_`) fail returning `'Unable to open object (message type not found)'` at `dset = dfile[ full_field_path ]` in `find_dataset(dfile, field_path)` function

**Comment by ax3l** [comment]
Thank you for the testing and great report!

for the 3rd and 4th point: Rémi will create a file here from FBPIC that I can use to debug 3 and he will look into 4.

**Comment by RemiLehe** [comment]
Here is the corresponding file
[Uploading diags.zip…]()

**Comment by RemiLehe** [comment]
@ax3l Here is a smaller reproducer for 3rd point in the original message by @hightower8083 (i.e. `openpmd-api` being unable to read FBPIC files):
```
import openpmd_api as io
io.Series('./diags/hdf5/data%T.h5', io.Access.read_only )
```
(This uses the files that I posted as a zip file.)

In this case, I get the following error:
```
HDF5-DIAG: Error detected in HDF5 (1.10.5) thread 0:
  #000: H5A.c line 425 in H5Aopen(): unable to load attribute info from object header for attribute: 'value'
    major: Attribute
    minor: Can't open object
  #001: H5Aint.c line 433 in H5A__open(): unable to load attribute info from object header for attribute: 'value'
    major: Attribute
    minor: Can't open object
  #002: H5Oattribute.c line 515 in H5O__attr_open_by_name(): can't locate attribute: 'value'
    major: Attribute
    minor: Object not found
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-2-3328cf1da57f> in <module>
----> 1 io.Series('./diags/hdf5/data%T.h5', io.Access.read_only )

[...]

PR #297, opened 2020-12-10
Fix bug in refactored reader when reading scalar fields

RemiLehe:
The refactored reader (#289) is crashing when the user passes a non-`None` `coord` for a scalar field, e.g.:
```
ts.get_field( field='rho', coord='x')
```
instead of 
```
ts.get_field( field='rho' )    # here, by default coord=None
```

While passing a non-`None` value for `coord` is here somewhat inconsistent, the previous reader (before #289) was robust to this and was naturally ignoring `coord` in the case of scalar fields. 
This PR recovers this previous behavior.


RemiLehe:
Note: this fixes part of #296

ax3l (approved):

ax3l:
Thank you for this fix! :sparkles:

PR #304, opened 2021-01-21
Add method to calculate central energy and energy spread

soerenjalas:
Hi,
I've added a method to calculate the central energy and the energy spread of a particle distribution via mean and standard deviation or median and median absolute deviation.
This is in part redundant with `get_mean_gamma` which I left for backward compatibility.


RemiLehe (changes requested):
Thanks a lot for this PR! 
I added a few comments for readability.

RemiLehe:
Thanks for updating the PR! ✨

RemiLehe (approved):

PR #305, opened 2021-02-12
Add support for fields with cylindrical geometry in DataReader

AngelFP:
Currently the `read_field_circ` method of the `DataReader` fails for fields with cylindrical geometry in both the `openpmd-api` and `h5py` backends. This is because the method expects a field with `thetaMode` geometry, which has a shape with three components where the first index contains the different modes. This is however not present in cylindrical data, which has a purely 2D shape. In addition, the is no ranking for cylindrical geometry in `get_grid_parameters`.

This PR fixes these issues in both backends, allowing the `DataReader` to work properly with this geometry. This has been tested in a new branch of VisualPIC, which uses the `DataReader` to access openPMD data, and everything seems to be working properly. The default tests also pass with no problem.

It is possible that there might be other parts of the code where data in cylindrical geometry could cause issues. I have seen for example that in the `lpa_diagnostics.py` or `interactive.py` the code checks for the geometry in some methods. However, I am only using the `DataReader`, so I have not tested for any problems outside of it.

Edit: sample diagnostics in cylindrical geometry. Download [here](https://desycloud.desy.de/index.php/s/egkbx69QG298mpN). Note that the fist iteration does not have fields.

### DIFF
--- a/openpmd_viewer/openpmd_timeseries/data_reader/h5py_reader/field_reader.py
+++ b/openpmd_viewer/openpmd_timeseries/data_reader/h5py_reader/field_reader.py
@@ -122,7 +122,7 @@ def read_field_circ( filename, iteration, field, coord,
                      slice_relative_position, slice_across, m=0, theta=0. ):
     """
     Extract a given field from an HDF5 file in the openPMD format,
-    when the geometry is thetaMode
+    when the geometry is thetaMode or cylindrical
 
     Parameters
     ----------
--- a/openpmd_viewer/openpmd_timeseries/data_reader/h5py_reader/field_reader.py
+++ b/openpmd_viewer/openpmd_timeseries/data_reader/h5py_reader/field_reader.py
@@ -176,8 +176,16 @@ def read_field_circ( filename, iteration, field, coord,
         field_path = join_infile_path( field, coord )
     group, dset = find_dataset( dfile, iteration, field_path )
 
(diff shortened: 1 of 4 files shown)


RemiLehe:
Thanks for this PR.
However, for the moment, the `cylindrical` geometry is not yet properly defined in the [`openPMD-standard`](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md). Therefore, I would be a bit reluctant to implement it in `openPMD-viewer`.

As an alterative, would it possible for the code that you are using to write the fields data as a `thetaMode` geometry (with only `m=0`)? If I am not mistaken, it would not take more memory: the data would simply be represented as an array of shape `(1, Nr, Nz)` instead of an array of shape `(Nr, Nz)`. And that would then make the data fully compliant with the openPMD standard.

Otherwise, another possibility would be to add the `cylindrical` geometry in next version of openPMD-standard (version 2.0.0).

AngelFP:
Hi Remi,

Thanks for the feedback. I wasn't aware that `cylindrical` geometry is not yet well defined in the standard. For the moment, the workaround you mention could solve the problem, so I'll look into it. It would be great however if this geometry could be added in the 2.0.0 standard. I would be happy to contribute if needed.

Should I close this PR for now?

RemiLehe:
OK, thanks a lot for being flexible!
I'll close the PR for now.

PR #306, opened 2021-02-12
Add list of available components to field metadata

AngelFP:
Small PR which adds the list of available field components (`'avail_components'`) to the field metadata. Right now, as far as I know, this can only be guessed from the field geometry, but it is not a given that all possible components are actually saved to file. As usual, this is needed by VisualPIC.


RemiLehe (approved):

RemiLehe:
Thanks for this PR!

PR #307, opened 2021-02-15
Add option to limit resolution of 3D reconstruction of thetaMode fields

AngelFP:
This PR proposes adding a parameter to `read_field_circ` for limiting the resolution of the 3D reconstructed fields when `theta=None`. Currently, the longitudinal and transverse resolution of the 3D reconstruction is the same as in the thetaMode fields, leading to very large memory usage and reconstruction time when using high-resolution data. This can be quite impractical for 3D data visualization.

Here, a `max_resolution_3d` parameter is added which allows the user to specify a list (e.g. `max_resolution_3d=[200, 100]`) containing the maximum longitudinal and transverse resolution that the 3D field should have, allowing for a significant performance gain.


RemiLehe (approved):
Thanks for this PR, and sorry for the delay!

PR #311, opened 2021-03-16
3d from circ interp

hightower8083:
presently the method `construct_3d_from_circ` gets field value at `(x_i, y_i)` from the _upper_ cell node `r_i`. 
In some case this may provide a lower accuracy, e.g. in some particular cases the full integral of laser `Ex**2` field could differ by up to a few percents from its real value (calculated on the actual cylindrical grid). 

This PR adds a linear interpolation to this method and increases the accuracy. For example, in the default example `lwfa_script.py` modified with `Nr = 350` and `Nz = 1800` the discrepancy in the total laser energy estimate is decreased from 0.054% to 0.002%.


RemiLehe (approved):

RemiLehe:
Looks good! Thanks a lot for this PR!

PR #313, opened 2021-03-25
openPMD-api reader: fix avail_circ_modes

### LINKED ISSUE #310: openpmd-api backend forgets modes m>1
Hi,
this might a bit of an edge case issue but it caused a bit of confusion for me so I thought it might be good to document. I have the case where I load a timeseries (fbpic) in a jupyter notebook via `d = LpaDiagnostics('lab_diags/hdf5', backend='openpmd-api')`,  and retrieve the fields with `d.get_field()`. Now I changed reran the simulation (going from 2 to 3 theta modes). After rerunning the cell to import the timeseries get_field() still returns the old fields now. With `backend='h5py'` I get the new data as expected. 
Is there some sort of caching going on with the openpmd-api that needs to be cleared?

ax3l:
Fix a mismatch between number of dimensions of a dataset and the extend of the first dimension (for modes).

Fix #310 

Thank you to @soerenjalas for reporting this! :tada:

### DIFF
--- a/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/params_reader.py
+++ b/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/params_reader.py
@@ -93,7 +93,7 @@ def read_openPMD_params(series, iteration, extract_parameters=True):
             if metadata['geometry'] == "thetaMode":
                 # simply check first record component
                 field_component = next(field.items())[1]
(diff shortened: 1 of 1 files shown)


RemiLehe (approved):

RemiLehe:
Thanks for fixing this @ax3l !

PR #315, opened 2021-04-07
[WIP] Extend API so as to extract the real momentum `p`

### LINKED ISSUE #314: momentum w/o mass
Just documenting this so we don't forget.

At the moment, we cannot read or visualize the momentum (`u(x|y|z)`) record of particle species that do not write `mass` as well.

```
openpmd_viewer/openpmd_timeseries/data_reader/io_reader/particle_reader.py in read_species_data(series, iteration, species_name, component_name, extensions)
     95         if np.all( m != 0 ):
     96             norm_factor = 1. / (m * constants.c)
---> 97             data *= norm_factor
     98 
     99 
ValueError: operands could not be broadcast together with shapes (265489,) (0,) (265489,) 
```

Related:
- #309 
- https://github.com/openPMD/openPMD-example-datasets/pull/14#issuecomment-815312809

RemiLehe:
Fix #314

### DIFF
--- a/openpmd_viewer/openpmd_timeseries/data_reader/h5py_reader/params_reader.py
+++ b/openpmd_viewer/openpmd_timeseries/data_reader/h5py_reader/params_reader.py
@@ -212,12 +212,15 @@ def simplify_record(record_comps):
     if 'momentum/x' in record_comps:
         record_comps.remove('momentum/x')
         record_comps.append('ux')
+        record_comps.append('px')
     if 'momentum/y' in record_comps:
         record_comps.remove('momentum/y')
         record_comps.append('uy')
+        record_comps.append('py')
     if 'momentum/z' in record_comps:
         record_comps.remove('momentum/z')
         record_comps.append('uz')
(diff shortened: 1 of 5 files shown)


openpmd_viewer/openpmd_timeseries/data_reader/h5py_reader/params_reader.py
```diff
@@ -212,12 +212,15 @@ def simplify_record(record_comps):
     if 'momentum/x' in record_comps:
         record_comps.remove('momentum/x')
         record_comps.append('ux')
```
ax3l, line 214: One could even go so far to only add "u(x|y|z)" aliases if the ED-PIC extension is set in the data series.

kevinCassou:
Hello 
Is there any chance to get this feature merged soon for output following the openPMD standard without a key "Mass" in the particle species group ?
Note that the "mass" does not appear in the standard, so it should be cool to read with the openPMD-viewer too! 
Many thanks

PR #318, opened 2021-06-09
CI: Fail On Error (Multiline)

ax3l:
Make sure that multi-line commands also fail on the first error that is occurring.
https://stackoverflow.com/questions/9952177/whats-the-meaning-of-the-parameter-e-for-bash-shell-command-line

Otherwise we might falsely continue execution.

Seen in https://github.com/openPMD/openPMD-api/pull/1004

More background:
- https://github.com/conda-incubator/setup-miniconda#important
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

### DIFF
--- a/.github/workflows/unix.yml
+++ b/.github/workflows/unix.yml
@@ -32,7 +32,7 @@ jobs:
         channels: conda-forge,defaults
         channel-priority: true
 
-    - shell: bash -l {0}
+    - shell: bash -e -l {0}
       name: Install dependencies
       run: |
         if [ "${{ matrix.python-version }}" != "3.8" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/unix.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         channels: conda-forge,defaults
         channel-priority: true
 
-    - shell: bash -l {0}
+    - shell: bash -e -l {0}
```
ax3l, line 35:
```suggestion
    - shell: bash -eo pipefail -l {0}
```

.github/workflows/unix.yml
```diff
@@ -41,9 +41,9 @@ jobs:
           conda install --yes cython numpy scipy h5py matplotlib jupyter pytest pyflakes python=${{ matrix.python-version }} python-wget 
         fi
 
-    - shell: bash -l {0}
+    - shell: bash -e  -l {0}
```
ax3l, line 44:
```suggestion
    - shell: bash -eo pipefail  -l {0}
```

.github/workflows/unix.yml
```diff
@@ -41,9 +41,9 @@ jobs:
...
       name: pyflakes
       run: python -m pyflakes openpmd_viewer
-    - shell: bash -l {0}
+    - shell: bash -e -l {0}
```
ax3l, line 47:
```suggestion
    - shell: bash -eo pipefail -l {0}
```

RemiLehe:
Thanks!

PR #321, opened 2021-09-17
setup.py: Python 3.6 - 3.10

ax3l:
We increased testing since 3.5 was EOL already a while ago, good to also update the meta data for PyPI et al.


setup.py
```diff
@@ -45,7 +45,7 @@ def run_tests(self):
         },
       cmdclass={'test': PyTest},
       platforms='any',
-      python_requires='>=3.5',
+      python_requires='>=3.6',
       classifiers=[
           'Programming Language :: Python',
           'Development Status :: 4 - Beta',
```
ax3l, line 51: @RemiLehe what do you think about changing that one as well? :)
```suggestion
          'Development Status :: 5 - Production/Stable',
```

setup.py
```diff
@@ -57,8 +57,8 @@ def run_tests(self):
           'Topic :: Scientific/Engineering :: Visualization',
           'Topic :: Database :: Front-Ends',
           'Programming Language :: Python :: 3',
-          'Programming Language :: Python :: 3.5',
           'Programming Language :: Python :: 3.6',
           'Programming Language :: Python :: 3.7',
-          'Programming Language :: Python :: 3.8'],
+          'Programming Language :: Python :: 3.8',
+          'Programming Language :: Python :: 3.9'],
```
ax3l, line 63: Python 3.10 is out :)
```suggestion
          'Programming Language :: Python :: 3.9',
          'Programming Language :: Python :: 3.10'],
```

PR #332, opened 2022-02-17
Fix MR Reads (ADIOS)

ax3l:
Start to fix MR reads on higher levels, where we have declared but undefined regions and need to filter by valid boxes.
### Old Implementation

Full read included UB regions that should not be read (standard: https://github.com/openPMD/openPMD-standard/pull/252).
After multiple reads, this could look like this:
Screenshot from 2022-02-17 12-24-04

### New Implementation

Mask invalid regions to zero, read only valid regions:
Screenshot from 2022-02-17 12-23-56

### DIFF
--- a/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
+++ b/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
@@ -11,6 +11,15 @@ License: 3-Clause-BSD-LBNL
 import numpy as np
 
 
+def chunk_to_slice(chunk):
+    """
+    Convert an openPMD_api.ChunkInfo to np.s_
+    """
+    stops = [a + b for a, b in zip(chunk.offset, chunk.extent)]
+    indices_per_dim = zip(chunk.offset, stops)
+    index_tuple = map(lambda s: slice(s[0], s[1], None), indices_per_dim)
+    return tuple(index_tuple)
+
 def get_data(series, record_component, i_slice=None, pos_slice=None,
              output_type=np.float64):
     """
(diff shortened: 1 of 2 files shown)


openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
```diff
@@ -60,8 +78,33 @@ def get_data(series, record_component, i_slice=None, pos_slice=None,
             list_index[dir_index] = i_slice[count]
         # Convert list_index into a tuple
         tuple_index = tuple(list_index)
-        # Slice dset according to tuple_index
-        data = record_component[tuple_index]
+        print("tuple_index={}".format(tuple_index))
+
```
ax3l, line 82: Slicing is still to do, a draft is written here.

openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
```diff
@@ -46,8 +55,17 @@ def get_data(series, record_component, i_slice=None, pos_slice=None,
...
     if pos_slice is None:
-        data = record_component[()]
+        # mask invalid regions with zero
+        data = np.zeros_like(record_component)
+        for chunk in chunks:
+            chunk_slice = chunk_to_slice(chunk)
+            # read only valid region
+            x = record_component[chunk_slice]
+            series.flush()
+            data[chunk_slice] = x
```
ax3l, line 68: We could use `load_chunk` to avoid the extra copy.

ax3l:
cc @tmsclark and @lucafedeli88 for review :)

RemiLehe (approved):

PR #334, opened 2022-03-04
MR Reads: Zero->NaN Padding

ax3l:
Fill invalid field regions in MR with `np.nan` instead of zeros.

@RemiLehe on Slack:
> matplotlib does handle `np.nan` gracefully:
> - For colormaps, it just shows white color
> - For line plots, it interrupts the curve at place where `nan` occurs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ openpmd_viewer.egg-info/
 .cache/
 .eggs/
 .vs/
+__pycache__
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@ openpmd_viewer.egg-info/
 .cache/
 .eggs/
 .vs/
+__pycache__
```
ax3l, line 10: Added because they keep showing up in my git status

ax3l:
@RemiLehe on Slack:
> Sounds good. matplotlib does handle `np.nan` gracefully:
> - For colormaps, it just shows white color
> - For line plots, it interrupts the curve at place where `nan` occurs.

RemiLehe:
This is working great:

- 2D plots with `get_field` and `plot=True`:
Screen Shot 2022-03-04 at 2 07 33 PM

- 1D lineouts with `get_field` and `plot=True`:
Screen Shot 2022-03-04 at 2 08 43 PM

ax3l:
Yay!! :star_struck:

PR #345, opened 2022-08-05
Allow calculation of waist for arbitrary propagation direction

RemiLehe:
The function `get_laser_waist` explicitly assumes that the `z` axis is the last axis.

This PR generalizes this code so that the `openPMD` checks which axis is actually to propagation axis.
It also allows the user to specify a different propagation axis than the `z` axis.

### DIFF
--- a/openpmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/openpmd_viewer/addons/pic/lpa_diagnostics.py
@@ -933,9 +933,9 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
         i_max = np.unravel_index( np.argmax( field ), field.shape )
         # Pick the corresponding transverse slice
         # (Transverse to laser propagation)
-        trans_slice = np.take( field, [i_max[slicing_index]], axis=slicing_index )
+        trans_slice = np.take( field, [i_max[slicing_index]], axis=slicing_index ).flatten()
         # Get transverse positions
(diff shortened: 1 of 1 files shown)


ax3l (approved):
This looks great, thank you.

Did you have a chance to try this also with a 2D and 3D data set?

ax3l:
@RemiLehe ping - ready besides a question on dimensionality

RemiLehe:
@ax3l Thanks for the reminder!
I now tested this PR with WarpX in 2D, RZ and 3D:
[inputs_2d.txt](https://github.com/openPMD/openPMD-viewer/files/11181682/inputs_2d.txt)
[inputs_3d.txt](https://github.com/openPMD/openPMD-viewer/files/11181683/inputs_3d.txt)
[inputs_rz.txt](https://github.com/openPMD/openPMD-viewer/files/11181684/inputs_rz.txt)
and here is the code that was used to test it:
```
import numpy as np
from openpmd_viewer.addons import LpaDiagnostics

ts = LpaDiagnostics('./diags_2d/diag1/')
print(ts.get_laser_waist(pol='y', iteration=200))

ts = LpaDiagnostics('./diags_RZ/diag1/')
print(ts.get_laser_waist(pol='y', iteration=10))

ts = LpaDiagnostics('./diags_3d/diag1/')
print(ts.get_laser_waist(pol='y', iteration=100))
```

In all cases, `get_laser_waist` returns the correct value (roughly 10 microns).

PR #347, opened 2022-09-01
When passing `t`, choose the closest iteration

RemiLehe:
When the user passes `t` in `get_particle` or `get_field`, `openPMD-viewer` used to choose the **last iteration before `t`**.

However, this is usually not what the user desires. Instead, `openPMD-viewer` should choose the iteration whose time is **closest to t**.

This PR fixes this problem.

### DIFF
--- a/openpmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/openpmd_viewer/addons/pic/lpa_diagnostics.py
@@ -63,7 +63,7 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
         ----------
         t : float (in seconds), optional
             Time at which to obtain the data (if this does not correspond to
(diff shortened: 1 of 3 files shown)


ax3l:
Just checking: this is now in sync with the logic in
https://github.com/openPMD/openPMD-viewer/blob/1.4.0/openpmd_viewer/openpmd_timeseries/interactive.py#L272-L278
?

I somehow remember we had an option to control the metric/algo for "closest" time... cannot spot it, maybe I misremembered and we did not have that in openPMD-viewer yet.

ax3l (approved):
Thanks, I think that's a good change.

Are there any user-facing doc-strings that need updating?

ax3l:
@RemiLehe ping - ready besides doc string?

PR #353, opened 2023-01-31
Allow openPMD-viewer to extract and plot complex numbers

RemiLehe:
Before this PR, `openPMD-viewer` was casting complex datasets to reals, and therefore discarding the imaginary part. 

Similarly, using `plt.plot` or `plt.imshow` with a complex array raises a warning, saying that the imaginary part was discarded. Here we choose to plot the modulus of the complex numbers instead.

This will be useful in order to plot the laser envelope, which is complex by nature.

### DIFF
--- a/openpmd_viewer/openpmd_timeseries/plotter.py
+++ b/openpmd_viewer/openpmd_timeseries/plotter.py
@@ -294,19 +294,22 @@ class Plotter(object):
         iteration = self.iterations[current_i]
         time = self.t[current_i]
 
-        # Get the title and labels
-        plt.title("%s at %.2e s   (iteration %d)" % (field_label,
-                    time, iteration), fontsize=self.fontsize)
-
-        # Add the name of the axes
-        plt.xlabel('$%s \;(m)$' % info.axes[0], fontsize=self.fontsize)
         # Get the x axis
         xaxis = getattr( info, info.axes[0] )
         # Plot the data
         if F.dtype in [np.complex256, np.complex128, np.complex64]:
             plot_data = abs(F) # For complex numbers, plot the absolute value
+            title = "|%s|" %field_label
         else:
             plot_data = F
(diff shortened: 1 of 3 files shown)


openpmd_viewer/openpmd_timeseries/plotter.py
```diff
@@ -303,7 +303,11 @@ def show_field_1d( self, F, info, field_label, current_i, plot_range,
         # Get the x axis
         xaxis = getattr( info, info.axes[0] )
         # Plot the data
-        plt.plot( xaxis, F )
+        if F.dtype in [np.complex256, np.complex128, np.complex64]:
+            plot_data = abs(F) # For complex numbers, plot the absolute value
```
MaxThevenet, line 307: Should we print a message?
```suggestion
            print('As the field is complex, its absolute value is plotted')
            plot_data = abs(F) # For complex numbers, plot the absolute value
```

MaxThevenet (approved):
Thanks! This looks good to me. It should fix the following issue, right? When doing
```python
ts = OpenPMDTimeSeries('.')
F, m = ts.get_field(field='laserEnvelope')
```
One gets
```
openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py:70: ComplexWarning: Casting complex values to real discards the imaginary part
  data = data.astype( output_type )
```

openpmd_viewer/openpmd_timeseries/plotter.py
ax3l: I think making this visible is a good idea, but a print might be very verbose in scripted/looped logic.

Could we maybe visible label the absolute, e.g., with `|F|` for complex fields in the plot?

RemiLehe: Good idea ; I just implemented this in the last commit.

PR #357, opened 2023-02-03
Grouped and overridable default values for slider widgets

hightower8083:
In general, when working with the slider, one often needs to check many boxes and fine-tune the plots. At the same time re-initialize of time-series requires re-initialization of slider and discards its setup. So it can happen that when one does quick checks on the simulation state during its progress via slider,  the slider state (widget values) has to be reset manually. 

This PR offers a simple and less invasive solution to fix and manipulate default widget values. Example of usage:
```python
# get a dictionary with default values
from openpmd_viewer.openpmd_timeseries.interactive import default_plot_profile as plot_profile

#override some values of choice
plot_profile['FieldColorbarMin'] = -5e6     # still have to check-activate the range
plot_profile['FieldColorbarMax'] = 5e6
plot_profile['ParticlesAlwaysRefresh'] = False
plot_profile['FieldAlwaysRefresh'] = False

# slider with new defaults
ts.slider(plot_profile=plot_profile)
```
Even without the access to override option , I think its in general more clear to have the initial widget values grouped in the code rather then distributed.


ax3l:
This is really cool, great idea.

If @RemiLehe agrees to the current approach, then we should add this to an example, e.g., `tutorials/3_Introduction-to-the-GUI.ipynb`, so people find it :)

PR #358, opened 2023-02-08
Avoid error in LpaDiagnostic functions when particle selection is empty

soerenjalas:
Some functions of the `lpa_diagnostic` add-on are causing errors when `get_particle` returns empty lists. This PR implements checks for this and has the functions return NaN in that case.


RemiLehe (approved):
Thanks for this PR!

PR #359, opened 2023-02-16
Improve get_laser_waist

soerenjalas:
Currently, when there are empty fields passed to `get_laser_waist` the method raises an exception. With this PR it returns a `NaN`instead.

Additionally this PR adds the option to calculate the waist from the projected field instead of a slice at the laser peak which can sometimes be a bit unreliable especially when looking at time series of the waist. This can be set via the `profile_method` argument.  This keeps the old way of getting the profile as default and shouldn't break previous script.


ax3l:
CI fix in #364

ax3l:
restarting CI

PR #363, opened 2023-03-07
Particle loading: Apply unitSI scaling factor to all types, not only floats

franzpoeschel:
In PIConGPU, the positionOffsets are integer types, so the particle histograms show wrong data without this patch.

### DIFF
--- a/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
+++ b/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
@@ -146,10 +146,12 @@ def get_data(series, record_component, i_slice=None, pos_slice=None,
     if (output_type is not None) and (data.dtype != output_type):
         data = data.astype( output_type )
     # Scale by the conversion factor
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thanks a lot!

As a background: PIConGPU stores the 3D cell index into `positionOffset` as integer. The `unitSI` is the cell size.

ax3l:
CI fix in #364

ax3l:
Restarting CI

openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
```diff
@@ -146,10 +146,12 @@ def get_data(series, record_component, i_slice=None, pos_slice=None,
     if (output_type is not None) and (data.dtype != output_type):
```
ax3l, line 146: Is it possible we should do this conversion after the scaling?

Current version:
- flexible, will not truncate, but ignores `output_type` for `unit_SI` != 1.0

Convert later:
- honor `output_type` but potentially crop precision

PR #372, opened 2023-05-18
Adding attributes to the FieldMetaInformation object

juliettepech:
This commit aims to add the simulation time and iteration to the info object returned by the get_field function.

 - [x] `h5py` io backend tested
 - [x] `openPMD-api` io backend tested

### DIFF
--- a/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/field_reader.py
+++ b/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/field_reader.py
@@ -16,7 +16,7 @@ from openpmd_viewer.openpmd_timeseries.field_metainfo import FieldMetaInformatio
 from openpmd_viewer.openpmd_timeseries.utilities import construct_3d_from_circ
 
 
-def read_field_cartesian( series, iteration, iterations, field_name, component_name,
+def read_field_cartesian( series, iteration, field_name, component_name,
                           axis_labels, slice_relative_position, slice_across ):
     """
     Extract a given field from a file in the openPMD format,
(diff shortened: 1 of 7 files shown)


openpmd_viewer/openpmd_timeseries/field_metainfo.py
```diff
@@ -12,6 +12,9 @@
 
 import numpy as np
 
+class OpenPMDException(Exception):
+    "Exception raised for invalid use of the openPMD-viewer API"
+    pass
```
ax3l: Seems to be a leftover :)
```suggestion
```

ax3l (commented):
Thank you, this looks great!

openpmd_viewer/openpmd_timeseries/data_reader/io_reader/field_reader.py
```diff
@@ -30,6 +30,10 @@ def read_field_cartesian( series, iteration, field_name, component_name,
     iteration: integer
         Iteration from which parameters should be extracted
 
+    iterations : array
+            An array of integers which correspond to the iteration of each file
+            (in sorted order)
+
```
ax3l: Probably left over from an earlier approach, but not used as a parameter now:
```suggestion
```

ax3l (approved):
Thank you very much for this! Looks great! :sparkles: 

I let @RemiLehe take a 2nd look.

ax3l:
Confirmed (and lazy uses `t` for the temporal axis, so it will not collide with the auto-generated `info.x/y/z/t` properties)

PR #378, opened 2023-06-22
setup.py: no upper version of openPMD-api

ax3l:
We break very seldomly and found this can confuse pip in complex scenarios.

Also, upper limits are considered bad practice, so we make it simpler now.


RemiLehe (approved):

RemiLehe:
Thanks!

PR #383, opened 2023-08-09
Impose that user always passes iteration or t

RemiLehe:
When the user does not pass `t` or `iteration`, the default behavior was to reuse the last iteration that was accessed. This can sometimes result in behavior that the user would not expected.

Instead, this PR imposes that `t` or `iteration` always be specified.


roelof-groenewald:
Thanks for this change!

ax3l (approved):

openpmd_viewer/openpmd_timeseries/main.py
```diff
@@ -630,7 +630,9 @@ def _find_output(self, t, iteration):
                       "The requested iteration '%s' is not available.\nThe "
                       "available iterations are: \n - %s\n" % (iteration, iter_list))
         else:
-            pass  # self._current_i retains its previous value
+            raise OpenPMDException(
+                "Please pass either a time (`t`) \nor an "
+                "iteration (`iteration`).")
```
ax3l: I would probably print this in a single line:
```suggestion
                "Please pass either a time (`t`) or an "
                "iteration (`iteration`).")
```

.github/workflows/unix.yml
```diff
@@ -38,12 +38,15 @@ jobs:
...
-          conda install --yes cython numpy scipy h5py matplotlib jupyter pytest pyflakes python=${{ matrix.python-version }} python-wget 
+          conda install --yes cython numpy scipy h5py matplotlib jupyter pytest pyflakes python=${{ matrix.python-version }} python-wget
         fi
 
     - shell: bash -eo pipefail  -l {0}
       name: pyflakes
-      run: python -m pyflakes openpmd_viewer
+      run: |
+        if [ "${{ matrix.python-version }}" != "3.7" ]; then
+          python -m pyflakes openpmd_viewer
```
ax3l: This can be rebased out: #385

PR #385, opened 2023-08-09
Do not use pyflakes for version 3.7

RemiLehe:
It seems that `pyflakes` dropped support for Python 3.7:
https://github.com/PyCQA/pyflakes/pull/752/files
and in practice, trying to use `pyflakes` with Python 3.7 indeed raises errors.

Therefore, this PR removes the `pyflakes` test for this version of Python.


ax3l (approved):
Thanks!

We should generally drop tests and support for Python 3.7, since it is EOL since last month:
https://endoflife.date/python

PR #390, opened 2023-08-16
Add all record attributes to `FieldMetaInformation`

AngelFP:
This PR proposes adding two new optional attributes to the `FieldMetaInformation` class. The new attributes, called `component_attrs` and `field_attrs` are dictionaries that contain all of the `openPMD` attributes of the field and its component. Until now, the `FieldMetaInformation` only extracts and stores a subset of this information, so if a field record contains attributes that are not explicitly read by the viewer, they cannot be accessed by the user.


RemiLehe (approved):
Thanks!

Issue #391, opened 2023-08-18
Momentum from .bp data


cbontoiu:
Hi,
I have a problem logged for PIConGPU here
https://github.com/ComputationalRadiationPhysics/picongpu/issues/4650

I have a model for which none of ux, uy,uz can be retrieved from the .bp data. The remaining variables, x, y,z and w can be obtained. I used openpmd viewer to read data from the .bp archives. There is nothing wrong with my installation because it can get ux, uy,uz from data saved a few months ago, so therefore with an older dev release of PIConGPU. Something happened in between.

In my speciesDefinition.param file I always have
using DefaultParticleAttributes = MakeSeq_t< position<position_pic>, momentum, weighting, particleId >;

This is the error I get: 

```
uz = ts_particles.get_particle( var_list=['uz'], iteration=ITER[nnRef], species='e', plot=True)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[37], line 1
----> 1 uz = ts_particles.get_particle( var_list=['uz'], iteration=ITER[nnRef], species='e', plot=True)

File ~/anaconda3/lib/python3.10/site-packages/openpmd_viewer/openpmd_timeseries/main.py:274, in OpenPMDTimeSeries.get_particle(self, var_list, species, t, iteration, select, plot, nbins, plot_range, use_field_mesh, histogram_deposition, **kw)
    272 data_list = []
    273 for quantity in var_list:
--> 274     data_list.append( self.data_reader.read_species_data(
    275         iteration, species, quantity, self.extensions))
    276 # Apply selection if needed
    277 if isinstance( select, dict ):

File ~/anaconda3/lib/python3.10/site-packages/openpmd_viewer/openpmd_timeseries/data_reader/data_reader.py:293, in DataReader.read_species_data(self, iteration, species, record_comp, extensions)
    290     return h5py_reader.read_species_data(
    291             filename, iteration, species, record_comp, extensions )
    292 elif self.backend == 'openpmd-api':
--> 293     return io_reader.read_species_data(
    294             self.series, iteration, species, record_comp, extensions )

File ~/anaconda3/lib/python3.10/site-packages/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/particle_reader.py:83, in read_species_data(series, iteration, species_name, component_name, extensions)
     81     if (macro_weighted == 1) and (weighting_power != 0):
     82         w_component = next(species['weighting'].items())[1]
[...]

PR #393, opened 2023-09-05
fix get_data for weighting in momentum reader

PrometheusPi:
In case of the ED-PIC extension, momentum values need to take into account the weighting. This failed for ADIOS/openPMD-api data - see issue #391. 
The reason for that was a missing argument in `getData( )`. This pull request fixes this issue.  

cc'ing @ax3l @cbontoiu


PrometheusPi:
@RemiLehe and @ax3l I am not sure what causes the CI to fail. Error message just reads some http time out. Might be not related to this pull request.

RemiLehe (approved):

ax3l:
Thank you for the fix!

PR #407, opened 2023-12-19
use `%matplotlib widget` by default

BenWibking:
This is necessary even with Jupyter Notebook, not just Jupyterlab. See comment: https://github.com/openPMD/openPMD-api/issues/1572#issuecomment-1862616267.

The matplotlib documentation suggests this rendering backend will work in all modern versions of Jupyter Notebook and Jupyterlab: https://matplotlib.org/stable/users/explain/figure/interactive.html#jupyter-notebooks-jupyterlab.

Fixes https://github.com/openPMD/openPMD-api/issues/1572.

PR #414, opened 2024-01-04
More `%matplotlib widget` in comments

ax3l:
Update leftover locations.
Fix broken RTD builds.

Follow-up to @BenWibking's #407


ax3l:
Merging to unbreak RTD

PR #340, opened 2022-04-12
Fix: Avoid `_like` function in Chunking

ax3l:
When we prepare chunked reads, we assume a single chunk for all backends but ADIOS2.
Preparing the returned data, we use `data = np.full_like(record_component, np.nan)`. It turns out that numpy seems to trigger a `__getitem__` access or full copy of our `record_component` at this point, which causes severe slowdown.

Refs.:
- https://github.com/numpy/numpy/blob/v1.22.0/numpy/core/numeric.py#L357-L418
- https://github.com/numpy/numpy/blob/v1.22.0/numpy/core/multiarray.py#L80-L145

This was first seen for particles, but affects every read where we do not slice a subset.

Thanks a lot to @AlexanderSinn for debugging this with me :sparkles: 

This fixes the performance regression that @AlexanderSinn, @MaxThevenet and @SeverinDiederichs saw in https://github.com/Hi-PACE/hipace/pull/725

Regression to #332 #334

PR #419, opened 2024-03-13
[WIP] Fill Invalid Regions w/ Zero

ax3l:
Filling unwritten/invalid regions, e.g., access outside of written ADIOS blocks, with not-a-number (NaN) created some confusion: https://github.com/ECP-WarpX/WarpX/issues/4775

Counter-point to this PR:
Technically, data in this region is literally not-a-number (yet). Matplotlib gracefully ignores NaNs for plotting, and can be configured to mask one way or the other with color. Padding with zeros is ambiguous: maybe data was not written, maybe data was written and truly zero.

We now always fill with zeros, hoping that "why is a region empty" ("because it was not written") does not create another kind of confusion.

Side note: resizing during write of BTD is non-trivial (as in: not supported by HDF5 and ADIOS), because we need to resize from the "lower index end". HDF5 and ADIOS only support resizing the upper index bound.

X-ref: #334 #340

### DIFF
--- a/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
+++ b/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/utilities.py
@@ -59,15 +59,14 @@ def get_data(series, record_component, i_slice=None, pos_slice=None,
     # ADIOS2: Actual chunks, all other: one chunk
     chunks = record_component.available_chunks()
 
-    # mask invalid regions with NaN: fill value
-    # note: NaN is only defined for floating point types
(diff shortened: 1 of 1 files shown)


ax3l:
@RemiLehe @jlvay note that we return NaNs often where technically sensible. E.g., LPA diagnostics and particle tracker return NaN if a particle selection does not apply (for a requested index).

Maybe this is just a lack of _documentation_.

ax3l:
Preferred solution for BTD:
Crop away the NaN regions and resize the meta-data to what was truly written.

Might require an extra openPMD meta-data attribute to track.

PR #431, opened 2024-10-23
Add calculation of EM energy

RemiLehe:
It is often useful to compute the total electromagnetic energy in the box. (This is oftentimes a good proxy for the laser energy.) This PR adds this function in `LpaDiagnostic`.

### DIFF
--- a/openpmd_viewer/addons/pic/lpa_diagnostics.py
+++ b/openpmd_viewer/addons/pic/lpa_diagnostics.py
@@ -702,13 +702,21 @@ class LpaDiagnostics( OpenPMDTimeSeries ):
         --------
         A float with the total electromagnetic energy inside the box (in Joules)
         """
+        # Check that the required data is available
+        if 'E' not in self.avail_fields or 'B' not in self.avail_fields:
+            raise ValueError('The fields E and B are required to compute the electromagnetic energy.')
+        # Check that the fields have either the thetaMode or 3dcartesian geometry
+        if (self.fields_metadata['E']['geometry'] not in ['thetaMode', '3dcartesian']) or \
+              (self.fields_metadata['B']['geometry'] not in ['thetaMode', '3dcartesian']):
+            raise ValueError('The electromagnetic energy can only be computed for 3D Cartesian and cylindrical simulations.')
+
         # For RZ: use `theta=None` to get the full 3D field
         Ex, info = self.get_field('E', 'x', t=t, iteration=iteration, theta=None )
-        Ey, info = self.get_field('E', 'y', m=1, iteration=iteration, theta=None )
-        Ez, info = self.get_field('E', 'z', m=1, iteration=iteration, theta=None )
-        Bx, info = self.get_field('B', 'x', m=1, iteration=iteration, theta=None )
-        By, info = self.get_field('B', 'y', m=1, iteration=iteration, theta=None )
-        Bz, info = self.get_field('B', 'z', m=1, iteration=iteration, theta=None )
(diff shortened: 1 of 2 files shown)


openpmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -675,6 +675,56 @@ def get_laser_envelope( self, t=None, iteration=None, pol=None,
...
+        Ey, info = self.get_field('E', 'y', t=t, iteration=iteration, theta=None )
+        Ez, info = self.get_field('E', 'z', t=t, iteration=iteration, theta=None )
+        Bx, info = self.get_field('B', 'x', t=t, iteration=iteration, theta=None )
+        By, info = self.get_field('B', 'y', t=t, iteration=iteration, theta=None )
+        Bz, info = self.get_field('B', 'z', t=t, iteration=iteration, theta=None )
+
+        # Compute the energy
+        energy_density = const.epsilon_0/2.*(Ex**2 + Ey**2 + Ez**2) + 1./(2*const.mu_0)*(Bx**2 + By**2 + Bz**2)
+        volume = info.dx*info.dy*info.dz # Cell volume
+        E = (energy_density*volume).sum()
```
dpgrote: A small optimization is to do the sum first, avoiding multiplying the whole array be the volume. Or is the volume an array also, the volume of each grid cell?
```
E = energy_density.sum()*volume
```

openpmd_viewer/addons/pic/lpa_diagnostics.py
```diff
@@ -675,6 +675,56 @@ def get_laser_envelope( self, t=None, iteration=None, pol=None,
...
+        --------
+        A float with the total electromagnetic energy inside the box (in Joules)
+        """
+        # Check that the required data is available
+        if 'E' not in self.avail_fields or 'B' not in self.avail_fields:
+            raise ValueError('The fields E and B are required to compute the electromagnetic energy.')
+        # Check that the fields have either the thetaMode or 3dcartesian geometry
+        if (self.fields_metadata['E']['geometry'] not in ['thetaMode', '3dcartesian']) or \
+              (self.fields_metadata['B']['geometry'] not in ['thetaMode', '3dcartesian']):
+            raise ValueError('The electromagnetic energy can only be computed for 3D Cartesian and cylindrical simulations.')
```
dpgrote, line 711: It looks like this should work for 2D and 1D, though depending on how info.dy and info.dz are defined.

RemiLehe: Yes, this is a bit tricky. `info.dy` does not exist in 2D and 1D in `openPMD-viewer`. 
Technically, we could multiply only by `info.dx` (instead of `info.dx` and `info.dy`) and get a result in J/m (instead of J), but I wanted to avoid this, so that the user does not get confused.

openpmd_viewer/addons/pic/lpa_diagnostics.py
RemiLehe:
```suggestion
        E = energy_density.sum() * volume
```

openpmd_viewer/addons/pic/lpa_diagnostics.py
dpgrote: For comparison, this is what is done in the WarpX reduced field energy diagnostic. The volume is calculated with the assumption that the extra dimensions have a length of 1. This would be easy to implement here so might as well for completeness so you wouldn't have to come back and submit another change if this is requested in the future.

RemiLehe: OK, sounds good. Will do.

PR #432, opened 2024-11-05
Fix figure display, when using %matplotlib widget

RemiLehe:
In general, using `%matplotlib widget` or `%matplotlib ipympl` should be almost equivalent (see https://github.com/matplotlib/ipympl/issues/529). However, when using `openPMD-viewer`, the slider works fine when doing `%matplotlib ipympl`, but does not when doing `%matplotlib widget` (i.e. the figures are not displayed, when doing `%matplotlib widget).

This PR fixes the issue.


dpgrote (approved):
Looks good.

PR #442, opened 2026-01-28
Added mpi reader 

guj:
Functions:
1.  distribute workload to ranks available 
2. gather back to rank 0 
3.  broadcast to all ranks
   note that when using Dask-mpi, often rank 1 is the working rank. Thus it is necessary to broadcast out
4. Jupyter users are likely to set plot=True.  For this feature it is limited to rank 0 to do plotting to avoid multiple plots on the notebook.


guj:
@ax3l  tests failed because tqdm is required in some notebooks but pytest does not have it in the env.

